PR #15, opened 2020-07-16
ZENKO-2675 Account creation page - 1st iteration

nicolas2bert:
UI Implementation of account creation page.

**Inputs**:
- Name
- Email
- Quotas (GB)

**API**: management API

**Call**: createConfigurationOverlayUser()[](url)

Screen Shot 2020-07-16 at 3 14 09 PM

### DIFF
--- a/eve/workers/keycloakconfig/keycloak-realm.json
+++ b/eve/workers/keycloakconfig/keycloak-realm.json
@@ -16,7 +16,7 @@
     "lastName" : "your_last_name",
     "email" : "sampleaccount1@sampling.com",
     "attributes" : {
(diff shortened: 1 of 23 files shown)


rahulreddy (commented):
Have you thought about if/how we can match the validation that Vault has for account names, account ids, email etc, for example here https://github.com/scality/Vault/blob/development/8.2/lib/Check.js#L107
Another thing that I realized is that, we recently implemented set account id feature where the user can specify their own account id, we should add the field during account creation, not in this PR but you can follow up later.

ilkescality (approved):
Just left few minor comments; all around LGTM :)

anurag4DSB (approved):

PR #22, opened 2020-08-12
ZENKO-2724 Accounts list: performance optimization

nicolas2bert:
Our application might have to render long lists of accounts (up to 1000). 

It is recommended to use a technique known as “windowing”. 

This technique only renders a small subset of the rows at any given time, and can dramatically reduce the time it takes to re-render the components as well as the number of DOM nodes created.


ilkescality (approved):

anurag4DSB (approved):

nicolas2bert:
@bert-e approve

PR #26, opened 2020-08-21
ZENKO-2742 manage empty state

nicolas2bert:
Screen Shot 2020-08-21 at 3 17 42 PM


src/react/account/Accounts.jsx
```diff
@@ -1,15 +1,49 @@
...
+    const accountList = useMemo(() => sortByDate(accounts, false), [accounts]);
+    const accountIndex = useMemo(() => accountList.findIndex(a => a.userName === accountNameParam), [accountList, accountNameParam]);
+
+    // empty state.
+    if (accountList.length === 0) {
+        return <L.Container>
+            <L.EmptyStateSection>
+                <Warning
+                    iconClass="fas fa-5x fa-wallet"
+                    title='Let&apos;s start, create your first account.'
```
wabernat: This is a run-on sentence. 

OK: "Let's start. Create your first account."

Better: "Create your first account."

If they will ever see this a second time: "Create your account."

anurag4DSB:
How do we login to the UI? Using super admin?
I can see you are logged in with your user (maybe via key cloak).
Just curious how the admin credentials are seeded.

nicolas2bert:
@anurag4DSB yes, "Zenko UI user" logs in through a client (RP) that supports OAuth 2.0 and relies on the OpenID provider.
An authorization code grant type with PKCE is used to obtain access token, ID token and refresh token.
So the authentication should work with any IDP that supports Authorization Code Flow with Proof Key for Code Exchange (PKCE).

anurag4DSB (approved):

ilkescality (approved):

src/react/account/Accounts.jsx
miniscruff: On an sort of related note, do we have plans to separate text into something more appropriate for localization? I can't imagine we don't have a requirement for supporting more than English.

nicolas2bert:
@bert-e approve

PR #36, opened 2020-10-01
ZENKO-2834 Upload object

nicolas2bert:
Screen-Recording-2020-10-01-at-12 20 57-PM

### DIFF
--- a/src/js/S3Client.js
+++ b/src/js/S3Client.js
@@ -1,6 +1,11 @@
 import AWS from 'aws-sdk';
 const async = require('async');
 
+const MULTIPART_UPLOAD = {
+    partSize: 1024 * 1024 * 6,
+    queueSize: 1,
+};
+
 export default class S3Client {
     constructor(params) {
         this.client = new AWS.S3({
@@ -80,4 +85,18 @@ export default class S3Client {
         return this.client.putObject(params).promise();
     }
 
+    uploadObject(bucketName, prefixWithSlash, files) {
+        return Promise.all(files.map(file => {
+            const key = `${prefixWithSlash}${file.name}`;
+            const params = {
+                Bucket: bucketName,
+                Key: key,
+                Body: file,
+                ContentType: file.type,
+            };
+            const options = { partSize: MULTIPART_UPLOAD.partSize,
+                queueSize: MULTIPART_UPLOAD.queueSize };
+            return this.client.upload(params, options).promise();
+        }));
+    }
 }
--- a/src/react/databrowser/objects/ObjectList.jsx
+++ b/src/react/databrowser/objects/ObjectList.jsx
@@ -4,11 +4,10 @@ import MemoRow, { createItemData } from './ObjectRow';
 import type { Object, S3Object } from '../../../types/s3';
 import React, { useMemo, useRef } from 'react';
 import Table, * as T from '../../ui-elements/Table';
-import { useFilters, useSortBy, useTable } from 'react-table';
+import { openFolderCreateModal, openObjectUploadModal } from '../../actions';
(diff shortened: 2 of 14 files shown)


src/js/S3Client.js
```diff
@@ -80,4 +85,19 @@ export default class S3Client {
         return this.client.putObject(params).promise();
     }
 
+    uploadObject(bucketName, prefixWithSlash, files) {
+        return new Promise((resolve, reject) =>
+            async.each(files, (file, callback) => {
```
miniscruff: I would just use promise.all instead of the async package. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all it works the same as far as I can tell.

src/react/databrowser/objects/ObjectList.jsx
```diff
@@ -74,11 +77,11 @@ export default function ObjectList({ objects, bucketNameParam, prefixParam }: Pr
         disableSortRemove: true,
         autoResetFilters: false,
         autoResetSortBy: false,
-    }, useFilters, useSortBy);
+    }, useFilters, useSortBy, useFlexLayout);
 
     return <L.ListSection>
         <T.ButtonContainer>
-            <T.ExtraButton icon={<i className="fas fa-upload" />} text="Upload" variant='info' onClick={() => dispatch(push(`/buckets/${bucketNameParam}/upload-object`))} size="default" type="submit" />
+            <T.ExtraButton icon={<i className="fas fa-upload" />} text="Upload" variant='info' onClick={() => dispatch(openObjectUploadModal())} size="default" type="submit" />
```
miniscruff, line 84: Just a personal preference here really but I find if you split your attributes on new lines it makes reviewing side by side easier and viewing partial diffs simpler as it would highlight the specific change and not the entire element. Following a line length rule anyway, this helps. ( Roughly 100 characters per line is my max )

Example:
```suggestion
            <T.ExtraButton
                icon={<i className="fas fa-upload" />}
                text="Upload"
                variant='info'
                onClick={() => dispatch(openObjectUploadModal())}
                size="default"
                type="submit"
            />
```

src/react/databrowser/objects/ObjectUpload.jsx
```diff
@@ -0,0 +1,157 @@
...
+};
+const ObjectUpload = ({ bucketName, prefixWithSlash }: Props) => {
+    const [acceptedFiles, setAcceptedFiles] = useState([]);
+    const [fileRejections, setFileRejections] = useState([]);
+    const show = useSelector((state: AppState) => state.uiObjects.showObjectUpload);
+    const dispatch: DispatchAPI<Action> = useDispatch();
+
+    const onDrop = (accepted, rejections) => {
+        if (accepted.length > 0) {
+            setAcceptedFiles([...acceptedFiles, ...accepted.filter(a => !acceptedFiles.find(f => f.path === a.path))]);
```
miniscruff: What if you filtered before setting the new files? Is there a chance that the filter clears all items in the array making no change? Like, even if you did drop some accepted items if they were already in the list this would do nothing? I guess it does nothing each time but if I remember correctly react will handle a few things when calling the set state methods so avoiding that if we can would be good?

nicolas2bert: Yes, we should avoid setting objects that are different by reference but equal by value to prevent unnecessary rerendering.

src/js/S3Client.js
nicolas2bert: It is cleaner, thank you :)

miniscruff (approved):

terrydervaux (approved):

terrydervaux:
LGTM for a first iteration.

nicolas2bert:
@bert-e approve

PR #38, opened 2020-10-05
ZENKO-2845 Delete files

nicolas2bert:
Screen-Recording-2020-10-05-at-11 47 02-AM

### DIFF
--- a/src/react/databrowser/objects/ObjectDelete.jsx
+++ b/src/react/databrowser/objects/ObjectDelete.jsx
@@ -1,0 +1,105 @@
+// @flow
+import React, { useMemo } from 'react';
+import Table, * as T from '../../ui-elements/Table';
+import { addTrailingSlash, formatBytes, maybePluralize } from '../../utils';
+import { closeObjectDeleteModal, deleteFiles, toggleAllObjects } from '../../actions';
+import { useDispatch, useSelector } from 'react-redux';
+import type { Action } from '../../../types/actions';
+import type { AppState } from '../../../types/state';
+import { Button } from '@scality/core-ui';
+import type { DispatchAPI } from 'redux';
+import { List } from 'immutable';
+import { CustomModal as Modal } from '../../ui-elements/Modal';
+import styled from 'styled-components';
+
+const INFO_DELETE_FOLDER = 'Non-empty folder won\'t be deleted.';
+
+const Files = styled.div`
+    height: 250px;
+    width: 500px;
+    overflow-y: scroll;
+    margin: 10px 0px;
+    border: 1px solid ${props => props.theme.brand.borderLight};
+`;
+
+const title = files => {
+    const foldersSize = files.filter(f => f.isFolder).size;
+    const objectsSize = files.size - foldersSize;
+    return `Do you want to permanently delete the selected ${maybePluralize(objectsSize, 'object')} ` +
+    `and ${maybePluralize(foldersSize, 'folder')}?`;
+};
+
+const fileSizer = files => {
+    const total = files.reduce((accumulator, file) => file.size ? accumulator + file.size : accumulator, 0);
+    return formatBytes(total);
+};
+
+type Props = {
+    toggled: List<Object>,
+    prefixParam: ?string,
+    bucketName: string,
+};
+const ObjectDelete = ({ bucketName, toggled, prefixParam }: Props) => {
+    const prefixWithSlash = addTrailingSlash(prefixParam);
+    const show = useSelector((state: AppState) => state.uiObjects.showObjectDelete);
+    const totalSize = useMemo(() => fileSizer(toggled), [toggled]);
+    const dispatch: DispatchAPI<Action> = useDispatch();
+
+    if (!show) {
+        return null;
+    }
+
+    const cancel = () => {
+        dispatch(toggleAllObjects(false));
+        dispatch(closeObjectDeleteModal());
+    };
+
+    const deleteSelectedFiles = () => {
+        if (toggled.size === 0) {
+            return;
+        }
+        const objects = toggled.map(s => {
+            return {
+                Key: `${prefixWithSlash}${s.name}`,
+                // TODO: add version when implemented.
+            };
+        }).toArray();
+        dispatch(deleteFiles(bucketName, prefixWithSlash, objects));
+    };
+
+    return (
+        <Modal
+            id="object-delete"
+            close={cancel}
+            footer={
+                <div>
+                    <Button outlined onClick={cancel} size='small' text='Cancel'/>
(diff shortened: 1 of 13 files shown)


terrydervaux (approved):
The user flow LGTM. Can you align the delete button on the right of the component? 

Screenshot 2020-10-05 at 14 08 02

naren-rajendran (approved):

miniscruff:
Your video says the two non-empty folders won't be deleted, but then it looks like they are removed. Also, the warning says you are going to delete 5 objects and 2 folders but the 2 folders wouldn't be removed. Or I am missing something which is totally possible.

src/react/databrowser/objects/ObjectDelete.jsx
```diff
@@ -0,0 +1,106 @@
...
+import type { DispatchAPI } from 'redux';
+import { List } from 'immutable';
+import { CustomModal as Modal } from '../../ui-elements/Modal';
+import styled from 'styled-components';
+
+const INFO_DELETE_FOLDER = 'Non-empty folder won\'t be deleted.';
+
+const Files = styled.div`
+    height: 250px;
+    width 500px;
```
miniscruff: I have noticed in a few PR's so far ( maybe I should have asked sooner? ) but is it a good idea to use pixels here? Is there a minimum resolution we support or is this scaled somewhere magically in JS land I dont see? ( Those magic css frameworks... just magic. )

src/react/databrowser/objects/ObjectDelete.jsx
```diff
@@ -0,0 +1,106 @@
...
+    overflow-y: scroll;
+    margin: 10px 0px;
+    border: 1px solid ${props => props.theme.brand.borderLight};
+`;
+
+const title = files => {
+    const foldersSize = files.filter(f => f.isFolder).size;
+    const objectsSize = files.size - foldersSize;
+    const objectTitle = `${objectsSize} object${objectsSize > 1 ? 's' : ''}`;
```
miniscruff: I feel like a utility method to pluralize words would be nice. Granted I do also think an i18n package or tool will be required later on anyway.

src/react/databrowser/objects/ObjectDelete.jsx
```diff
@@ -0,0 +1,106 @@
...
+    const deleteSelectedFiles = () => {
+        if (toggled.size === 0) {
+            return;
+        }
+        const objects = toggled.map(s => {
+            return {
+                Key: `${prefixWithSlash}${s.name}`,
+                // TODO: add version when implemented.
+            };
+        }).toArray();
```
miniscruff, line 66: I thought `.map` already returned an array?

src/react/databrowser/objects/ObjectDelete.jsx
nicolas2bert: I want to control the width of this component, for now. 
The application should fit inside a 768px x 1024px display. (@terrydervaux)
So, setting the width to 500px should be fine. 
Also, I am open to any suggestions.

src/react/databrowser/objects/ObjectDelete.jsx
nicolas2bert: About internationalization, we are planning on implementing it since it is now required.

src/react/databrowser/objects/ObjectDelete.jsx
nicolas2bert: `toggled` is an immutable List and needs to be transformed into a javascript array before passing it to AWS.S3.deleteObjects method.

nicolas2bert:
For this first iteration, I don't check if folders are empty or not. 
I am just informing the user that "Non-empty folder won't be deleted."
Thank you for your review @miniscruff!

miniscruff:
> For this first iteration, I don't check if folders are empty or not.
> I am just informing the user that "Non-empty folder won't be deleted."
> Thank you for your review @miniscruff!

oh ok, could you maybe place that comment at the top or bottom and not per folder maybe? Cause it seems to be related to that folder when it may not be.

src/react/databrowser/objects/ObjectDelete.jsx
miniscruff: oh ok, so it wasn't already an array.

nicolas2bert:
@miniscruff yes, I took note here: https://scality.atlassian.net/browse/ZENKO-2865. 
I will change it in another iteration after talking to our UX expert: @dvasilas

src/react/databrowser/objects/ObjectDelete.jsx
miniscruff: ok 768 sounds like a wide minimum, I would say that is too wide for a mobile user given the pixel scale on phones. Also did you mean 1024x768 landscape? At my last job we used a 350px minimum width. So media queries or percentages were common for us. By no means was our css great but just want to mention 300 or 350px was our base case.

src/react/utils/index.js
```diff
@@ -40,3 +40,5 @@ export function formatBytes(bytes, decimals = 2) {
 
 export const stripTrailingSlash = name => name.slice(-1) === '/' ? name.slice(0, -1): name;
 export const addTrailingSlash = name => name ? name.slice(-1) !== '/' ? `${name}/`: name : '';
+
+export const maybePluralize = (count, noun, suffix = 's') => `${count} ${noun}${count > 1 ? suffix : ''}`;
```
miniscruff, line 44: haha I just want to say I like the function name.

src/react/databrowser/objects/ObjectDelete.jsx
```diff
@@ -0,0 +1,105 @@
...
+    overflow-y: scroll;
+    margin: 10px 0px;
+    border: 1px solid ${props => props.theme.brand.borderLight};
+`;
+
+const title = files => {
+    const foldersSize = files.filter(f => f.isFolder).size;
+    const objectsSize = files.size - foldersSize;
+    return `Do you want to permanently delete those ${maybePluralize(objectsSize, 'object')} ` +
+    `and ${maybePluralize(foldersSize, 'folder')}?`;
```
miniscruff: Do you want to permanently delete those object and folder?
Do you want to permanently delete those objects and folders?
Neither of these really sounds right to me but I am not a grammar expert.

Just an idea for your consideration:
Do you want to permanently delete the selected object(s) and folder(s)?

miniscruff (approved):

src/react/databrowser/objects/ObjectDelete.jsx
nicolas2bert: Thank you for your feedback, today we don't support mobile interface. 
I have taken note of your comments and will make changes if the requirements change.

nicolas2bert:
@bert-e approve

PR #40, opened 2020-10-06
ZENKO 2776 createBucket Tests

romainlancia:
[+] test BucketCreate component
[+] test createBucket action

### DIFF
--- a/src/react/actions/__tests__/s3bucket.test.jsx
+++ b/src/react/actions/__tests__/s3bucket.test.jsx
@@ -1,0 +1,49 @@
+import * as actions from '../s3bucket';
+import * as dispatchAction from './utils/dispatchActionsList';
+import {
+    OWNER_NAME,
+    errorS3State, initState, testDispatchFunction,
+} from './utils/testUtil';
+
+const createBucketNetworkAction = dispatchAction.NETWORK_START_ACTION('Creating bucket');
+const listBucketsNetworkAction = dispatchAction.NETWORK_START_ACTION('Listing buckets');
+
+describe('s3bucket actions', () => {
+    const asyncTests = [
+        {
+            it: 'createBucket: should return expected actions',
+            fn: actions.createBucket({
+                name: 'azeaze',
+                locationConstraint: {
+                    value: 'us-east-1',
+                },
+            }),
+            storeState: initState,
+            expectedActions: [
+                createBucketNetworkAction,
+                listBucketsNetworkAction,
+                dispatchAction.LIST_BUCKETS_SUCCESS_ACTION([], OWNER_NAME),
+                dispatchAction.NETWORK_END_ACTION,
+                dispatchAction.LOCATION_PUSH_ACTION('/buckets'),
+                dispatchAction.NETWORK_END_ACTION,
+            ],
+        },
+        {
+            it: 'createBucket: should handle error',
+            fn: actions.createBucket({
+                name: 'azeaze',
+                locationConstraint: {
+                    value: 'us-east-1',
+                },
+            }),
+            storeState: errorS3State(),
+            expectedActions: [
+                createBucketNetworkAction,
+                dispatchAction.HANDLE_ERROR_SPEC_ACTION('The server is temporarily unavailable.'),
+                dispatchAction.NETWORK_END_ACTION,
+            ],
+        },
+    ];
+
+    asyncTests.forEach(testDispatchFunction);
+});
--- a/src/react/actions/__tests__/utils/testUtil.js
+++ b/src/react/actions/__tests__/utils/testUtil.js
@@ -1,11 +1,12 @@
 // @flow
 import { ErrorMockManagementClient, account, latestOverlay, location } from '../../../../js/mock/managementClient';
+import { ErrorMockS3Client, ownerName } from '../../../../js/mock/S3Client';
 import { ErrorUserManager, MockUserManager } from '../../../../js/mock/userManager';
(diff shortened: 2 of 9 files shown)


src/react/actions/__tests__/s3bucket.test.jsx
```diff
@@ -0,0 +1,49 @@
...
+    authenticatedUserState, errorS3State,
+    mockStore, OWNER_NAME, testDispatchFunction,
+} from "./utils/testUtil";
+import * as dispatchAction from "./utils/dispatchActionsList";
+
+const createBucketNetworkAction = dispatchAction.NETWORK_START_ACTION('Creating bucket');
+const listBucketsNetworkAction = dispatchAction.NETWORK_START_ACTION('Listing buckets');
+
+describe('s3bucket actions', () => {
+    it('createBucket: should return expected actions', () => {
```
nicolas2bert: This test can be moved to asyncTest array.

src/react/actions/__tests__/utils/testUtil.js
```diff
@@ -70,6 +72,19 @@ export function errorManagementState(): AppState {
     };
 }
 
+export function errorS3State(): AppState {
+    const state = initState;
+    return {
+        ...state,
+        auth: {
```
nicolas2bert: Why do you need it?

src/react/databrowser/buckets/__tests__/BucketCreate.test.jsx
```diff
@@ -0,0 +1,107 @@
+import * as actions from '../../../actions';
+import BucketCreate from '../BucketCreate';
+import React from 'react';
+import { act } from 'react-dom/test-utils';
+import { reduxMount } from '../../../utils/test';
+
+// Jest requires a setter on the tested object but ES6 doesn't have
+// but jest.mock() allows you solving this by mocking your required module after the import.
+jest.mock('../../../actions', () => ({
```
nicolas2bert: Is it possible to mock `clearServerError` method from <BucketCreate> component. 
It might avoid us having to do this action mock.

src/react/databrowser/buckets/__tests__/BucketCreate.test.jsx
```diff
@@ -0,0 +1,107 @@
...
+            } else if (t.expectedMinLengthNameError !== null) {
+                expect(component.find('ErrorInput#error-name').text()).toContain(t.expectedMinLengthNameError);
+            } else if (t.expectedMaxLengthNameError !== null) {
+                expect(component.find('ErrorInput#error-name').text()).toContain(t.expectedMaxLengthNameError);
+            }
+
+        });
+    });
+
+    it('Should not call clearServerError when clicking inside the component', () => {
```
nicolas2bert: I'm not sure you are testing what you described.

nicolas2bert (changes requested):
First iteration.

src/react/databrowser/buckets/__tests__/BucketCreate.test.jsx
romainlancia: Ok so I did some research and  I think it is not possible to mock clearServerError method from the component for many reasons:
1) the fact thar CreateBucket is not a class, indeed in this issue https://github.com/enzymejs/enzyme/issues/944 someone manage to spyOn a function inside a component using instance() but his component is a class and according to the enzyme documentation we can't do an instance() on a functional component since react 16.x , you can check it here: https://enzymejs.github.io/enzyme/docs/api/ShallowWrapper/instance.html
2) the other way was to use dive() instead of instance() but this needs to use shallow instead of mount which is not possible as BucketCreate need the store to work but shallow doesn't render the entire tree only the first child

src/react/account/__tests__/Accounts.test.jsx
```diff
@@ -67,7 +67,7 @@ describe('Accounts', () => {
         const { component } = reduxMount(<MemoryRouter><Accounts/></MemoryRouter>, {
             configuration: {
                 latest: {
-                    users: [ account1, account2 ],
```
nicolas2bert: Is your linter yelling because of these extra spaces? I don't think it's mandatory.

src/react/databrowser/buckets/__tests__/BucketCreate.test.jsx
```diff
@@ -0,0 +1,115 @@
...
+        const component = await reduxMountAct(
+            <BucketCreate/>, {
+                uiErrors: {
+                    errorMsg: 'error',
+                    errorType: 'byComponent',
+                },
+            },
+        );
+
+        const cancelButton = component.findWhere(n => n.name() === 'Button' && n.prop('outlined') === true);
```
nicolas2bert: Feel free to add an id: `#cancel-btn` that will help you find the "cancel button" component. 
The component's props might change over time, hence make our tests fail. 
We should avoid relying on them.

src/react/account/__tests__/Accounts.test.jsx
romainlancia: No it didn't but I can definitely change this

src/react/databrowser/buckets/__tests__/BucketCreate.test.jsx
romainlancia: Ok I'll do that

nicolas2bert (approved):

terrydervaux (approved):
LGTM.

romainlancia:
@bert-e approve

PR #43, opened 2020-10-08
ZENKO-2859 Object Summary

nicolas2bert:
Add summary tab:
Screen-Recording-2020-10-09-at-7 06 31-PM

### DIFF
--- a/src/react/databrowser/objects/ObjectRow.jsx
+++ b/src/react/databrowser/objects/ObjectRow.jsx
@@ -2,7 +2,10 @@
 
 import * as T from '../../ui-elements/Table';
 import React, { memo } from 'react';
-import type { S3Bucket } from '../../../types/s3';
+import { toggleAllObjects, toggleObject } from '../../actions';
+import type { Action } from '../../../types/actions';
+import type { DispatchAPI } from 'redux';
+import type { Object } from '../../../types/s3';
 import { areEqual } from 'react-window';
 import isDeepEqual from 'lodash.isequal';
 import memoize from 'memoize-one';
@@ -11,7 +14,7 @@ type PrepareRow = (RowType) => void;
 
 type RowType = {
     id: number,
-    original: S3Bucket,
+    original: Object,
     cells: any,
     getRowProps: (any) => void,
 };
@@ -21,6 +24,7 @@ type RowsType = Array<RowType>;
 type Data = {
     rows: RowsType,
     prepareRow: PrepareRow,
+    dispatch: DispatchAPI<Action>,
 };
 
 type RowProps = {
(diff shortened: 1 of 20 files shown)


src/react/databrowser/objects/ObjectRow.jsx
```diff
@@ -35,21 +42,38 @@ type RowProps = {
...
     index,
     style,
 }: RowProps) => {
     const row = rows[index];
     prepareRow(row);
+    const isSelected = objectName && objectName === row.original.name;
+
+    const handleClick = () => {
+        if (row.original.isFolder || isSelected) {
+            return;
```
rachedbenmustapha: Would the MD pane need to be cleared to stay in sync, if an object was selected (displaying its MD on the right), then a folder was selected? Also if multiple objects are selected.

In S3 browser the selection system is kind of confusing to people because there is both the checkbox and the "row selection" that are not synchronized - how does the right pane handle this?

src/react/utils/index.js
```diff
@@ -42,3 +42,10 @@ export const stripTrailingSlash = name => name.slice(-1) === '/' ? name.slice(0,
 export const addTrailingSlash = name => name ? name.slice(-1) !== '/' ? `${name}/`: name : '';
 
 export const maybePluralize = (count, noun, suffix = 's') => `${count} ${noun}${count > 1 ? suffix : ''}`;
+
+export function stripQuotes(s) {
+    if (s.charAt(0) === '"' && s.charAt(s.length-1) === '"') {
```
miniscruff: not sure if its any faster but I prefer to use startsWith and endsWith like:
```suggestion
    if (s.startsWith('"') && s.endsWith('"')) {
```

src/react/databrowser/objects/ObjectRow.jsx
nicolas2bert: The way we designed it was:
- user can check (using checkboxes) multiple rows (objects or folders) and apply an action on then (ie deletion). I use the verb "toggle" for that kind of "selection" in the code. 

- user can select only one row (object only) at a time by clicking inside the Row. It will show the object metadata. I use the verb "select" for that in the code.

- User clicks on the Row name to either get into a folder or download an object. 


So, user can not select a folder but can toogle it and then apply the delete action on it. User is not able to select a folder and so won't see the folder metadata.

nicolas2bert: @terrydervaux @dvasilas We need to rethink the design since it is confusing to have the checkbox and the "row selection" that are not synchronized.

miniscruff (approved):

src/react/databrowser/objects/ObjectRow.jsx
rachedbenmustapha: I've had this feedback several times on orbit's browser (which is s3browser), because people are already used to selecting objects (emails in gmail, files in the Finder, aws s3's object browser, etc) and expect a certain behavior, which is generally different from the "highlighting a row is different than checking checkbox(es)" s3browser behavior with regards to what displays in the sidebar/statusbar and what is affected by actions on the selection.

nicolas2bert: - The checkbox and the "row selection" are now synchronized. 
- If a folder is selected, the MD pane shows: "Select an object."
- If multiple items (objects, folders, or both) are selected, the MD pane shows: "Multiple items selected."
I updated the video on the description.

miniscruff (approved):

rachedbenmustapha (approved):

nicolas2bert:
@bert-e approve

PR #47, opened 2020-10-15
ZENKO-2875 Slimming down the bundle(s)

nicolas2bert:
We manage to reduce the size of the bundle from 5.23MB to 2.55MB then 823KB after compression.

- Use a tool to analyze our bundle: `webpack-bundle-analyzer`. 
- “Smart” import: when using heavy libraries, make sure that we only import exactly what we need.
- Compress bundle to further reduce the bundle size.

- For testing purposes, we added: ZENKO-2817 Write Nginx proxy config for S3, IAM and STS services

TODO: Code splitting: the bundle is getting larger, UI takes a long time to load.
"code-splitting" our UI can help “lazy-load” just the things that are currently needed by the user.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /public/assets/js
 /public/assets/fonts
 /public/assets/index.html
+/public/assets/index.html.gz
 
 yarn-error.log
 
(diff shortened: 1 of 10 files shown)


miniscruff (approved):

jonathan-gramain (approved):

nicolas2bert:
@bert-e approve

PR #56, opened 2020-10-29
ZENKO-2903 Error creating user metadata

nicolas2bert:
`x-amz-meta-` prefix is set twice.

_Note: tests are coming._

### DIFF
--- a/src/react/databrowser/objects/details/Metadata.jsx
+++ b/src/react/databrowser/objects/details/Metadata.jsx
@@ -1,20 +1,16 @@
 // @flow
 import { AddButton, Buttons, Char, Footer, Header, HeaderKey, HeaderValue, InputExtraKey, InputValue, Inputs, Item, Items, SubButton } from '../../../ui-elements/EditableKeyValue';
 import { Button, Select } from '@scality/core-ui';
+import { META_WORD, defaultMetadataKeys, isEmptyItem } from '../../../utils';
 import type { MetadataItems, ObjectMetadata } from '../../../../types/s3';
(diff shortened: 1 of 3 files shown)


src/react/utils/index.js
```diff
@@ -1,3 +1,8 @@
+export const defaultMetadataKeys = ['cache-control', 'content-disposition', 'content-encoding',
```
miniscruff, line 1: If the only use is to check if a key is included I would use a set here.

src/react/utils/index.js
```diff
@@ -1,3 +1,8 @@
...
+    'content-language', 'content-type', 'expires', 'website-redirect-location'];
+
+export const META_WORD = 'x-amz-meta';
```
miniscruff, line 4: Small thing but I would use `AMZ_META` or similar as it would indicate its from amazon and not one of our own meta values. Plus its 2 words at least. :)

src/react/databrowser/objects/details/Metadata.jsx
```diff
@@ -38,7 +34,7 @@ const convertToAWSMetadata = (items: MetadataItems) => {
         }
         if (isMetaWord(items[i].key)) {
             const metaKey = items[i].metaKey || '';
-            result[`${items[i].key}-${metaKey}`] = items[i].value;
+            result[metaKey] = items[i].value;
```
miniscruff, line 37: This looks like a good unit test imo

nicolas2bert:
Closed because of bigger refacto.

PR #57, opened 2020-10-31
ZENKO-2903 Error creating user metadata

nicolas2bert:
We should differentiate "system metadata controlled by user" from "user metadata".
Test: more to come.

### DIFF
--- a/src/react/databrowser/objects/details/Metadata.jsx
+++ b/src/react/databrowser/objects/details/Metadata.jsx
@@ -1,20 +1,14 @@
 // @flow
+import { AMZ_META, METADATA_SYSTEM_TYPE, METADATA_USER_TYPE, isEmptyItem, systemMetadata } from '../../../utils';
 import { AddButton, Buttons, Char, Footer, Header, HeaderKey, HeaderValue, InputExtraKey, InputValue, Inputs, Item, Items, SubButton } from '../../../ui-elements/EditableKeyValue';
 import { Button, Select } from '@scality/core-ui';
-import type { MetadataItems, ObjectMetadata } from '../../../../types/s3';
+import type { MetadataItem, MetadataItems, ObjectMetadata } from '../../../../types/s3';
 import React, { useEffect, useMemo, useState } from 'react';
-import { isEmptyItem } from '../../../utils';
 import { putObjectMetadata } from '../../../actions';
 import styled from 'styled-components';
 import { useDispatch } from 'react-redux';
 
-const EMPTY_ITEM = { key: '', value: '' };
-
-const OPTIONS = ['cache-control', 'content-disposition', 'content-encoding',
-    'content-language', 'content-type', 'expires', 'website-redirect-location',
-    'x-amz-meta'];
-
-const META_WORD = 'x-amz-meta';
+const EMPTY_ITEM = { key: '', value: '', type: '' };
 
 const TableContainer = styled.div`
     overflow-y: auto;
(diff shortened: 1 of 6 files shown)


nicolas2bert:
@miniscruff I had to make some changes since we were not differentiating "system metadata controlled by user" from "user metadata".

src/react/utils/index.js
```diff
@@ -1,3 +1,18 @@
+export const systemMetadataKeys = ['CacheControl', 'ContentDisposition',
```
miniscruff: I would create this dynamically using something like `systemMetadata.map(d => d.key)`

src/react/databrowser/objects/details/Metadata.jsx
```diff
@@ -1,49 +1,49 @@
...
+        if (isUserMetadata(items[i].type)) {
+            userMetadata[items[i].key] = items[i].value;
+        }
+        if (isSystemMetadata(items[i].type)) {
+            systemMetadata[items[i].key] = items[i].value;
         }
     }
-    return result;
+    return { systemMetadata, userMetadata };
 };
```
miniscruff: A few ideas here, 1) I would pass in the metadata to the IsXMetadata methods as that is kinda what the name implies.
2) Use the for...of loop.

( Github doesn't let you suggest when there are deleted lines... )

```js
const isUserMetadata = metadata => metadata.type === METADATA_USER_TYPE;
const isSystemMetadata = metadata => meatadata.type === METADATA_SYSTEM_TYPE;

const convertToAWSMetadata = (items: MetadataItems) => {
    const userMetadata = {};
    const systemMetadata = {};
    for (let metadata of items) {
        if (isEmptyItem(metadata)) {
            continue;
        }
        if (isUserMetadata(metadata)) {
            userMetadata[metadata.key] = metadata.value;
        } else if (isSystemMetadata(metadata)) {
            systemMetadata[metadata.key] = metadata.value;
        }
    }
    return { systemMetadata, userMetadata };
};
```

Alternatively you could use a filter for empty items then a filter+map per type, not sure on performance but doing 5 things in a for loop compared to a for loop with 1 thing 5 times is generally comparable.

src/react/databrowser/objects/details/Metadata.jsx
```diff
@@ -112,8 +115,22 @@ function Properties({ objectMetadata }: Props) {
...
+        if (type) {
+            if (isSystemMetadata(type) && key) {
+                return selectOptions.find(l => l.value === key);
+            }
+            if (isUserMetadata(type)) {
+                return userMetadataOption;
+            }
+            return;
+        }
+        return;
```
miniscruff: Neither return is required here

src/react/databrowser/objects/details/Metadata.jsx
```diff
@@ -112,8 +115,22 @@ function Properties({ objectMetadata }: Props) {
...
+
+    const selectValue = (metadata: MetadataItem) => {
+        const { key, type } = metadata;
+        if (type) {
+            if (isSystemMetadata(type) && key) {
+                return selectOptions.find(l => l.value === key);
+            }
+            if (isUserMetadata(type)) {
+                return userMetadataOption;
+            }
```
miniscruff: Based on my suggestion above for the isXMetadata function this could look like:
```suggestion
        if (isSystemMetadata(metadata) && metadata.key) {
            return selectOptions.find(l => l.value === key);
        }
        if (isUserMetadata(metadata)) {
            return userMetadataOption;
        }
```

src/react/databrowser/objects/details/Metadata.jsx
miniscruff: For the alternative it would be something like:
Edit: you can not map from an array to an Object so you would have to either use Reduce OR the new fromEntries ( https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries )
```js
const nonEmpty = items.filter(isFilledItem) // I would rename isEmpty to the more positive isFilled or similar
return {
  systemMetadata: Object.fromEntries(nonEmpty.filter(isSystemMetadata).Map(item => [item.key, item.value])
  userMetadata: Object.fromEntries(nonEmpty.filter(isUserMetadata).Map(item => [item.key, item.value])
}
```

That kinda big one-liner could be a method as well but probably will go with the other anyway.

nicolas2bert: I decided to use `for of` and change `isXMetadata` to `isXType` for clarity purposes.

src/react/databrowser/objects/details/Metadata.jsx
```diff
@@ -26,21 +26,22 @@ const selectOptions = systemMetadata.map(o => ({
...
+    for (const item of items) {
+        const { type, key, value } = item;
+        if (isEmptyItem(item)) {
             continue;
         }
-        if (isUserMetadata(items[i].type)) {
-            userMetadata[items[i].key] = items[i].value;
+        if (isUserType(type)) {
+            userMetadata[key] = value;
         }
```
miniscruff: personally I would use else if here as it would seem to me that an item would be either user or system type and not both. Could just be me though.

src/react/databrowser/objects/details/Metadata.jsx
```diff
@@ -126,25 +140,25 @@ function Properties({ objectMetadata }: Props) {
                 <Items>
                     {
                         items.map((p, i) => {
-                            const isShrink = isMetaWord(p.key);
-                            return <Item isShrink={isShrink} key={i}>
+                            const isUserMD = isUserType(p.type);
+                            return <Item isShrink={isUserMD} key={i}>
```
miniscruff, line 144: is using the index for a react key ideal? wouldn't you want to use something that is unique despite order? Not sure on this but I remember always using some sort of value when I was using react.

naren-rajendran (approved):

src/react/databrowser/objects/details/Metadata.jsx
nicolas2bert: We don’t have "stable IDs" for those rendered items (something like p.id). 
Also, in this case, the order of items should not change so it should be fine to use indexes for keys.

miniscruff (approved):

romainlancia (approved):
LGTM

nicolas2bert:
@bert-e approve

PR #61, opened 2020-11-18
fix-console-errors-due-to-multiple-same-ids

romainlancia:


src/react/databrowser/objects/ObjectDelete.jsx
```diff
@@ -83,8 +83,8 @@ const ObjectDelete = ({ bucketName, toggled, prefixWithSlash }: Props) => {
                 <Table>
                     <T.Body>
                         {
-                            toggled.map(s => (
-                                <T.Row key={s.name}>
+                            toggled.map((s, index) => (
+                                <T.Row key={`${s.name}-${index}`}>
```
nicolas2bert: Why is this needed? `s.key` should be unique.

romainlancia: I saw a warning about a non unique key from this line

nicolas2bert: We should try to avoid `index` when we can. https://reactjs.org/docs/lists-and-keys.html#keys

nicolas2bert (approved):
Thank you!

dora-korpar (approved):

anurag4DSB (approved):

romainlancia:
@bert-e approve

PR #70, opened 2020-12-16
zenko-3050 release stage

tcarmet:
Experimental, but the idea here is to create a stage `tag` that we will use to perform all the steps required to release zenko-ui.

Ideally this stage would be triggered automatically when tagging the project, but we don't have this feature (yet). It doesn't stop us from being prepared.

I've created two namespaces in registry.scality.com
registry.scality.com/zenko-ui -> for tagged images, production usage
registry.scality.com/zenko-ui-dev -> for dev images

both namespaces are public to the world, as zenko-ui is a public repository. therefore we don't need credentials to pull the images, but need to push them, which the CI does it for us.

the `tag` stage, which can be forced manually, will take a tag as a paremeter, check it's revision, pull the image from the dev namespace and publish it into the production one.

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -39,6 +39,13 @@ models:
         docker build --pull -t ${SCALITY_OCI_REPO_DEV}:%(prop:commit_short_revision)s .;
       haltOnFailure: True
       env: *deploy-env
+  - ShellCommand: &docker-login
+      name: login to docker registry
+      command: >-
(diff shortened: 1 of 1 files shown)


nicolas2bert (approved):
Thank you Thomas! LGTM 🥇

terrydervaux (approved):
LGTM, thanks Thomas

PR #69, opened 2020-12-15
ZENKO-3043 Render more than 1000 objects/versions

nicolas2bert:
Screen-Recording-2020-12-15-at-5 17 02-PM

PR #72, opened 2020-12-22
ZENKO-3056 Fix unexpected error messages

nicolas2bert:
- List object: Decrease the number of keys returned in the response -> **reduce server response time**. 
- Infinite scroll: force to only load 1 page of items at a time.
- Wait for objects to be loaded before accessing the first object metadata (if it exists).
- Improve AWS API error messages that are rendered.

### DIFF
--- a/src/react/databrowser/objects/ObjectListTable.jsx
+++ b/src/react/databrowser/objects/ObjectListTable.jsx
@@ -1,6 +1,6 @@
 // @flow
 import MemoRow, { createItemData } from './ObjectRow';
-import React, { useMemo } from 'react';
+import React, { useCallback, useMemo } from 'react';
 import Table, * as T from '../../ui-elements/Table';
 import { continueListObjects, toggleAllObjects, toggleObject } from '../../actions';
 import { useDispatch, useSelector } from 'react-redux';
@@ -12,6 +12,7 @@ import InfiniteLoader from 'react-window-infinite-loader';
 import { List } from 'immutable';
 import type { Object } from '../../../types/s3';
 import { formatBytes } from '../../utils';
+import { push } from 'connected-react-router';
 import styled from 'styled-components';
 
 export const Icon = styled.i`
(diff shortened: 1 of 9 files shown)


src/react/databrowser/objects/ObjectListTable.jsx
```diff
@@ -35,11 +36,20 @@ type Props = {
 export default function ObjectListTable({ objects, bucketName, toggled, isVersioningType, prefixWithSlash }: Props){
     const dispatch = useDispatch();
     const nextMarker = useSelector((state: AppState) => state.s3.listObjectsResults.nextMarker);
-    const isItemLoaded = index => {
-        const shouldRefetch = nextMarker && index === objects.size - 1;
+    const loading = useSelector((state: AppState) => state.networkActivity.counter > 0);
```
JBWatenbergScality, line 39: in order to simplify a bit the application you can create a `useTypedSelector` like so : 
```js
export const useTypedSelector: <TSelected>(
  selector: (state: AppState) => TSelected,
  equalityFn?: (left: TSelected, right: TSelected) => boolean
) => TSelected = useSelector;
```

That you can then reuse everywhere without the necessity to type explicitly the state parameter

src/react/databrowser/objects/ObjectListTable.jsx
```diff
@@ -35,11 +36,20 @@ type Props = {
...
+    const objectsLength = objects.size;
+    const isToggledFull = toggled.size > 0 && toggled.size === objectsLength;
+
+    const isItemLoaded = useCallback(index => {
```
JBWatenbergScality, line 43: index parameter could be typed

JBWatenbergScality (approved):

src/react/databrowser/objects/ObjectListTable.jsx
nicolas2bert: `flowtype/require-parameter-type` has not been enforced by our linter yet.

src/react/databrowser/objects/ObjectListTable.jsx
nicolas2bert: Yes, I have a plan to do that using the helper type: `TypedUseSelectorHook`.

terrydervaux (approved):
LGTM 👍 

This fix seems to solve the issue reported in that comment: https://github.com/scality/zenko-ui/pull/69#pullrequestreview-554870406.

The navigation in the objects browsing pages is more fluid.

nicolas2bert:
@bert-e approve

nicolas2bert:
bert-e reset

PR #74, opened 2021-01-05
feature: ZENKO-3076 Add coverage reporting

miniscruff:
Adds code coverage report to readme and codecov upload

https://codecov.io/gh/scality/zenko-ui/ to view the web UI for coverage.
https://app.codecov.io/gh/scality/zenko-ui/branch/feature%2FZENKO-3076_AddCodeCoverage for this branch/PR.

Note: the badge in the readme will show unknown until this is merged as there isn't a build on the default branch yet.

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -19,6 +19,11 @@ models:
       KEYCLOAK_PASSWORD: "123"
       KEYCLOAK_USER_FIRSTNAME: "Bart"
       KEYCLOAK_USER_LASTNAME: "Simpson"
+  - env: &coverage
+      CODECOV_TOKEN: '%(secret:codecov_token)s'
+      VCS_COMMIT_ID: '%(prop:revision)s'
+      VCS_BRANCH_NAME: '%(prop:branch)s'
+      CI_BUILD_ID: '%(prop:bootstrap)s'
   - Git: &clone
       name: fetch source
       repourl: '%(prop:git_reference)s'
@@ -67,7 +72,7 @@ stages:
           haltOnFailure: True
       - ShellCommand:
           name: run test suite
-          command: yarn test
+          command: yarn test:coverage
           haltOnFailure: True
       - ShellCommand: *yarn-build
       - ShellCommand: *docker-build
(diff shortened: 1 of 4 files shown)


nicolas2bert (approved):
Thank you! Great new year gift!

naren-rajendran (approved):

eve/main.yml
```diff
@@ -88,6 +93,12 @@ stages:
           env:
             <<: *deploy-env
             <<: *keycloak-env
+      - ShellCommand:
+          name: upload test coverage
+          haltOnFailure: False
```
tcarmet, line 98: Just a suggestion, the default is already `haltOnFailure: False` so we could remove this line.

But sometimes you do want to make this explicit so I'll leave it up to you.
```suggestion
```

tmacro (approved):

eve/main.yml
```diff
@@ -88,6 +93,12 @@ stages:
           env:
             <<: *deploy-env
             <<: *keycloak-env
+      - ShellCommand:
+          name: upload test coverage
```
tcarmet, line 97: I wonder if coverage upload should "always run" even if the previous step have failed. I'm starting to think that we only want to upload successful results. But I could be missing something.

If we want to always run, I suggest we add the option

```suggestion
          name: upload test coverage
          alwaysRun: true
```
If not then ignore my comment.

tcarmet (approved):
Open suggestions/comments, but LGTM

alexanderchan-scality (approved):

eve/main.yml
miniscruff: Edit: nvm I got the default mixed up, I am going to keep this as I do want to mention it will not fail builds if the upload fails.

eve/main.yml
miniscruff: I would say no. The reason being, depending on which part of the build failed could result in incorrect coverage. An example is if we have a syntax error in some test that fails to run, we might miss out on otherwise tested code. Or if a build fails then the coverage is meaningless anyway.

PR #82, opened 2021-01-13
ZENKO-3042 Fix list versions toggle

nicolas2bert:
- "List Versions" toggle should be disabled if bucket versioning is “Disabled”. 
- "List Versions" toggle should be enabled if bucket versioning is “Suspended“ or “Enabled“.

### DIFF
--- a/src/react/databrowser/objects/ObjectList.jsx
+++ b/src/react/databrowser/objects/ObjectList.jsx
@@ -2,8 +2,9 @@
 
 import * as L from '../../ui-elements/ListLayout2';
 import * as T from '../../ui-elements/Table';
+import type { BucketInfo, ListObjectsType, Object } from '../../../types/s3';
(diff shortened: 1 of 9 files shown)


naren-rajendran (approved):

src/react/databrowser/objects/ObjectList.jsx
```diff
@@ -2,8 +2,9 @@
 
 import * as L from '../../ui-elements/ListLayout2';
 import * as T from '../../ui-elements/Table';
+import type { BucketInfo, ListObjectsType, Object } from '../../../types/s3';
```
anurag4DSB: I checked `types/s3` to see what `Object` is.

Is it recommended/intentional to use the reserved `Object` ?

nicolas2bert: It was not intentional. ZenkoUI has a compiler to strip away Flow types.

nicolas2bert: I will change it to ObjectEntity for readability.

anurag4DSB (approved):

nicolas2bert:
@bert-e approve

PR #85, opened 2021-02-16
ZENKO-3258 Use npm to compile core-UI

nicolas2bert:
With yarn, lifecycle scripts (e.g. `prepare` here:https://github.com/scality/core-ui/blob/development/1.0/package.json#L14) for hosted git dependencies was not running on install.
Swith from `yarn` to `npm` to avoid transpiling `core-ui` dependency.


eve/main.yml
```diff
@@ -29,13 +29,14 @@ models:
       repourl: '%(prop:git_reference)s'
       retryFetch: true
       haltOnFailure: true
-  - ShellCommand: &yarn-install
+  - ShellCommand: &npm-install
       name: install dependencies
-      command: yarn install --frozen-lockfile
+      # set unsafe-perm to true to run core-ui "prepare" script
+      command: npm config set unsafe-perm true && npm ci
```
JBWatenbergScality, line 35: You may want to install dependencies as part of https://github.com/scality/zenko-ui/blob/feature/ZENKO-3258/npm/eve/workers/build/Dockerfile to benefit from caching

nicolas2bert: Not sure to understand, when would you clone the project before installing its dependencies?

JBWatenbergScality: I meant we could do something similar to https://github.com/scality/metalk8s/blob/development/2.8/eve/workers/pod-unit-tests/ui/Dockerfile#L34 but not sure if it makes sense here as well

nicolas2bert: Ok, I see that change as a different "improvement". I will try to merge this one first and see if we can improve it later. Thanks for your comment.

JBWatenbergScality: Ok will give you my approval. LGTM !

JBWatenbergScality (approved):

ChengYanJin (approved):
Congrats to have the latest core-ui version ;)

nicolas2bert:
@bert-e approve

PR #95, opened 2021-03-26
ZENKO-3180: Update version of core-ui to v0.14.1

theocerutti:
Update core-ui version from v0.10.0 to v0.14.1.

PR #96, opened 2021-03-29
ZENKO-3313 account creation page discrepencies

theocerutti:
This pull request resolve:

ZENKO-3192
ZENKO-3193
ZENKO-3199
ZENKO-3200
ZENKO-3198
ZENKO-3314
ZENKO-3193: Tooltip background is not fully covering the text (it will be resolved next coreui update: https://github.com/scality/zenko-ui/pull/95)

### DIFF
--- a/src/react/ui-elements/FormLayout.jsx
+++ b/src/react/ui-elements/FormLayout.jsx
@@ -7,33 +7,38 @@ import type { Node } from 'react';
 import React from 'react';
 import styled from 'styled-components';
 
-// TEMPLATE:
-//
-// <Form>
-//     <F.Title> global title </F.Title>
-//     <F.Fieldset>
-//         <F.Label tooltipMessages={['text1', 'text2']> label1 </F.Label>
-//         <F.Input/>
-//         <F.ErrorInput hasError={false}> error Name </F.ErrorInput>
-//     </F.Fieldset>
-//     <F.Fieldset>
-//         <F.Label tooltipMessages={['text1', 'text2']}> label2 </F.Label>
-//         <F.Input />
-//         <F.ErrorInput hasError={false}> error Email </F.ErrorInput>
-//     </F.Fieldset>
-//     <F.Footer>
-//         <F.FooterError> <Banner> Global error </Banner> </F.FooterError>
-//         <F.FooterButtons> <Button/> <Button/> </F.FooterButtons>
-//     </F.Footer>
-// </Form>
+/* TEMPLATE:
+<FormContainer>
+    <F.Form>
+        <F.Title> global title </F.Title>
+        <F.Fieldset>
+            <F.Label tooltipMessages={['text1', 'text2']> label1 </F.Label>
+            <F.Input/>
+            <F.ErrorInput hasError={false}> error Name </F.ErrorInput>
+        </F.Fieldset>
+        <F.Fieldset>
+            <F.Label tooltipMessages={['text1', 'text2']}> label2 </F.Label>
+            <F.Input />
+            <F.ErrorInput hasError={false}> error Email </F.ErrorInput>
+        </F.Fieldset>
+        <F.Footer>
+            <F.FooterError> <Banner> Global error </Banner> </F.FooterError>
+            <F.FooterButtons> <Button/> <Button/> </F.FooterButtons>
+        </F.Footer>
+    </F.Form>
+    <F.Form>
+        // ...
+    </F.Form>
+</FormContainer>
+*/
 
 
 export const Title = styled.div`
     display: flex;
-
     text-transform: capitalize;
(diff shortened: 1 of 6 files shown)


Cuervino:
7ff3354 - Fix ZENKO-3192: remove uppercase Title, update font size to 24px
Great

febaddd - Update button forms color to blue (ZENKO-3199), Remove the Quota Area
Great

cfab059 - Update ZENKO-3200(extends form frame) + update test for quota

- Should be **top-centered,** not only centered.
image
- The frame should be extended, not removed.

cfab059 - Update ZENKO-3193: tooltip
Icon is good.

e2ed92c - Update ZENKO-3193 tooltipcontainer
The tooltip frame itself is buggy. Does that come from this PR?
image

08ce91e - Update color FormLayout Title 
Looks good

theocerutti:
> [cfab059](https://github.com/scality/zenko-ui/commit/cfab05917d47440747fa3a0ff3b442f25b2b65a8) - Update ZENKO-3200(extends form frame) + update test for quota
> 
> * Should be **top-centered,** not only centered.
>   image
> * The frame should be extended, not removed.

On my side, the commit cfab059 have the frame centered vertically/horizontally with the extended background, do you have the same problem with the last commit?

Here is the result:
image

> e2ed92c - Update ZENKO-3193 tooltipcontainer
> The tooltip frame itself is buggy. Does that come from this PR?

The tooltip is buggy but it is not because of this PR. We have to update the version of core-ui. (https://github.com/scality/zenko-ui/pull/95 should fixed the tooltip)

Cuervino:
> On my side, the commit cfab059 have the frame centered vertically/horizontally with the extended background, do you have the same problem with the last commit?

- My  bad for the background, I did not cherry-pick correctly this commit. The layout is good.

- However, for the "centering", it should be horizontally aligned (now it's good), but I don't like the Vertical alignment.
So either remove the vertical alignment (hence my "top alignment" comment), either add a 20% padding on the top for having a nice effect, in case of _small_ forms ("Account creation" for instance).

theocerutti:
I don't understand:

Do you want this alignment: (horizontally centered + padding on the top)
image

or this: (vertically centered and padding on the left)
image

Cuervino:
> 
> 
> I don't understand:
> 
> Do you want this alignment: (horizontally centered + padding on the top)
>
> 
> or this: (vertically centered and padding on the left)
>

The first one is the target :-)

Cuervino:
The form "Add New Storage Location" is the longest one, in case of some Location type that needs lot of fields.

In this case, the issue is that the frame/background is not until the bottom:
image

If it's not my branch issue, could you please fix this?

Cuervino (approved):
Great, Thanks, LGTM !

src/react/Routes.jsx
```diff
@@ -45,18 +46,20 @@ function Routes() {
                     ]}
                 />
             </NavbarContainer>
-            <Switch>
-                <Route exact path="/" render={() => <Redirect to="/accounts" />}/>
+            <L.Container>
```
nicolas2bert: What is this `L.Container` layer element used for?
We might want to remove it.
If needed, `ui-elements/ListingLayout2` components are specific to listing page, please use `ui-elements/Container` instead.

src/react/ui-elements/FormLayout.jsx
```diff
@@ -90,13 +94,18 @@ const LabelContainer = styled.label`
...
+    margin-left: 8px;
+`;
+
+const IconQuestionCircle = styled.i`
+    color: #434343;
 `;
 
 const UlOverlay = styled.ul`
-    width: 200px;
+    width: 200;
```
nicolas2bert: Omitting units? 
We might want to get rid of the `UlOverlay` styled component if core-ui handle the tooltip sizing.

theocerutti: Oh yes, my bad for the unit.
However I can't remove the UlOverlay because this is what allows us to have the bullet points before the tooltip message.
But yes, I have to remove the width to allow core-ui to resize the tooltip as its want.

Then without width rule in UlOverlay and the new version of CoreUI it will become like this:
image
image

src/react/Routes.jsx
theocerutti: Fixed, I found a way to remove L.Container without breaking changes.

nicolas2bert:
@theocerutti please squash commits and rebase (we might have to fix conflict) and we can merge it.

nicolas2bert (approved):

nicolas2bert:
@theocerutti your branch has to start with one of the following prefix: `feature/*, documentation/*, improvement/*, bugfix/*, w/*, q/*, hotfix/*` for the CI to run.

PR #97, opened 2021-04-07
RINGX-116 shell UI integration

nicolas2bert:
- Integrate shell-UI navbar (https://github.com/scality/metalk8s/tree/development/2.9/shell-ui)
- Clean up existing authentication logic.
- Add and adapt tests.


eve/main.yml
```diff
@@ -8,6 +8,8 @@ branches:
     stage: post-merge
 
 models:
+  - env: &shell-ui-env
+      SHELL_UI_IMAGE: 'registry.scality.com/playground/jbwatenberg/shell-ui:2.9-dev'
```
nicolas2bert, line 13: Need to be updated.

src/react/Navbar.js
```diff
@@ -0,0 +1,105 @@
...
+
+function InternalNavbar() {
+    const navbarEndpoint = useSelector((state: AppState) => state.auth.config.navbarEndpoint);
+    const navbarConfigUrl = useSelector((state: AppState) => state.auth.config.navbarConfigUrl);
+    useWebComponent(navbarEndpoint, 'solutions-navbar');
+
+    const navbarRef = useRef<NavbarWebComponent | null>(null);
+
+    useLoginEffect(navbarRef);
+
```
JBWatenbergScality, line 95: You may also want to fetch the theme from the navbar (https://github.com/scality/metalk8s/blob/development/2.9/ui/src/components/Navbar.js#L221)

src/react/Routes.jsx
```diff
@@ -1,67 +1,63 @@
...
-                <Route exact path="/create-location" component={LocationEditor} />
-                <Route path="/locations/:locationName/edit" component={LocationEditor} />
-
-                <Route path='/accounts/:accountName?' component={Accounts} />
-                <Route path="/create-account" component={AccountCreate} />
-
-                <Route path={['/buckets/:bucketName?', '/buckets/:bucketName/objects', '/create-bucket']} component={DataBrowser} />
-                <Route path={['/create-workflow', '/workflows/:ruleId?']} component={Workflows} />
-                <Route path="*" component={NoMatch}/>
+                <Route exact path="/login/callback" component={LoginCallback}/>
```
JBWatenbergScality: I'm not sure this login/callback is still usefull

src/react/auth/LoginCallback.jsx
```diff
@@ -1,18 +1,14 @@
...
+import React from 'react';
 
 function LoginCallback() {
-    const dispatch: DispatchAPI<Action> = useDispatch();
-
-    useEffect(() => {
-        dispatch(signinCallback());
-    });
+    const location = useLocation();
+    // AFTER processing response from the authorization endpoint, shell UI redirect to /login/callback
```
JBWatenbergScality: Not exactly, the navbar redirects to the page you were trying to access before triggering the OIDC danse. So removing the route /login/callback should be fine

src/react/reducers/oidc.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+import type { OIDCAction } from '../../types/actions';
+import type { OIDCState } from '../../types/state';
+
+import { initialNetworkActivityState } from './initialConstants';
+
+export default function oidc(state: OIDCState=initialNetworkActivityState, action: OIDCAction) {
+    switch (action.type) {
+    case 'ADD_OIDC_USER':
```
JBWatenbergScality, line 9: nit: this could be extracted to a constant to guarantee sync with `src/react/actions/oidc.js`

src/react/auth/LoginCallback.jsx
nicolas2bert: Are we talking about the same redirect? I was talking about this one: 
https://github.com/scality/metalk8s/blob/development/2.9/shell-ui/src/index.js#L158
https://github.com/bjerkio/oidc-react/blob/master/src/AuthContext.tsx#L112

src/react/Navbar.js
nicolas2bert: Fetching the theme from the navbar could be added in another PR, since this PR focuses on authentication.

src/react/reducers/oidc.js
nicolas2bert: Yes, thank you, I will create a ticket for that.

src/react/Routes.jsx
nicolas2bert: I have not found a better way to deal with the authorization server redirecting the user back to `ui.zenko.local` with an authorization code
Indeed, if I do not specify `redirectUrl`:
1) auth server will redirect to `http://ui.zenko.local/accounts/test-account-1?state=<STATE>&session_state=<SESSION_STATE>&code=<CODE>`. 
2) solutions-navbar--authenticated will fire
3) UI will load the clients and requests will fail (be canceled) because shell UI redirects to http://ui.zenko.local/accounts/test-account-1 (onSignIn is fired) https://github.com/scality/metalk8s/blob/development/2.9/shell-ui/src/index.js#L158

There is no UI visible errors (since the redirect step 3 clean errors up with the redirect) but it seems wrong to me to redirect to a private route. What do you think @JBWatenbergScality ?

JBWatenbergScality: My approach would be the following : 
- Use `http://ui.zenko.local/` as the redirectUrl
- When trying to access a private route while being unauthenticated (such as `/accounts/test-account-1`), the navbar will save this url and redirect the user to the auth server 
- Once authenticated, the auth server will redirect to `http://ui.zenko.local/?state=<STATE>&session_state=<SESSION_STATE>&code=<CODE>`
- The navbar will then redirect back to `http://ui.zenko.local/accounts/test-account-1` (https://github.com/scality/metalk8s/blob/development/2.9/shell-ui/src/index.js#L158 indeed)
- `solutions-navbar--authenticated` is triggered giving access to the protected route

nicolas2bert: The only issue I was facing is `Once authenticated, the auth server will redirect to http://ui.zenko.local/?state=<STATE>&session_state=<SESSION_STATE>&code=<CODE>` step, where my user was authenticated (userManager.events.addUserLoaded)  and clients (S3, API..) were initializing before being redirected by https://github.com/scality/metalk8s/blob/development/2.9/shell-ui/src/index.js#L158 which leads to some canceled calls.

JBWatenbergScality (approved):

alexanderchan-scality (approved):

nicolas2bert:
@bert-e approve

PR #98, opened 2021-04-08
ZENKO-3355 Update UX on replication GUI

theocerutti:
Hello!

> More vertical space between the header "Create New Rule" and "Rule Type" (24px)

There was already a 40px margin-bottom. Should I lower the margin to 24px?

>     Add the tooltips next to the label that needs explanation (+ Terry to provide the content)
waiting for Terry

### DIFF
--- a/src/react/workflow/WorkflowContent.jsx
+++ b/src/react/workflow/WorkflowContent.jsx
@@ -8,7 +8,6 @@ import { CustomTabs } from '../ui-elements/Tabs';
 import React from 'react';
 import Replication from './replication/Replication';
 import type { S3BucketList } from '../../types/s3';
-import { Title } from '../ui-elements/FormLayout';
 import { Warning } from '../ui-elements/Warning';
 import type { Workflow } from '../../types/workflow';
 import { push } from 'connected-react-router';
@@ -43,7 +42,8 @@ function WorkflowContent({ createMode, wfSelected, bucketList }: Props) {
                 <CreationSection>
                     <Table id=''>
                         <T.Body autoComplete='off'>
-                            <Title> create new Workflow </Title>
+                            <T.Title> Create New Workflow </T.Title>
+                            <T.Subtitle> All * are mandatory fields </T.Subtitle>
                             <Replication loading={loading} showEditWorkflowNotification={false} workflow={null} replications={replications} bucketList={bucketList} locations={locations} createMode={true} />
                         </T.Body>
                     </Table>
--- a/src/react/workflow/WorkflowList.jsx
+++ b/src/react/workflow/WorkflowList.jsx
@@ -6,8 +6,8 @@ import { AutoSizer } from 'react-virtualized';
 import { FixedSizeList } from 'react-window';
 import { ListSection } from '../ui-elements/ListLayout3';
 import React from 'react';
-import type { Workflows } from '../../types/workflow';
 import { TextTransformer } from '../ui-elements/Utility';
+import type { Workflows } from '../../types/workflow';
 import { push } from 'connected-react-router';
 import { useDispatch } from 'react-redux';
 
--- a/src/react/workflow/WorkflowList.jsx
+++ b/src/react/workflow/WorkflowList.jsx
@@ -7,6 +7,7 @@ import { FixedSizeList } from 'react-window';
 import { ListSection } from '../ui-elements/ListLayout3';
 import React from 'react';
 import type { Workflows } from '../../types/workflow';
+import { TextTransformer } from '../ui-elements/Utility';
 import { push } from 'connected-react-router';
 import { useDispatch } from 'react-redux';
 
@@ -19,12 +20,13 @@ const columns = [
     {
         Header: 'Workflow Type',
         accessor: 'type',
+        Cell({ value: type }: { value: string }) { return <TextTransformer transform='capitalize'>{type}</TextTransformer>; },
         width: 25,
     },
     {
         Header: 'State',
         accessor: 'state',
-        Cell: ({ value }) => { return value ? 'active' : 'inactive'; },
+        Cell: ({ value }) => { return value ? 'Active' : 'Inactive'; },
         width: 15,
     },
 ];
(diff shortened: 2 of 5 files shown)


Cuervino (changes requested):
> 1.         Change the string to "Create New Rule" (with the correct Pascal Case)
OK

> 2.         More vertical space between the header "Create New Rule" and "Rule Type" (24px)
Yes please make it 24 px (looks better)

> 3.         Add "All * are mandatory fields" between this title and "Rule Type", and add a "*" next to all the required fields ("Rule name", Source"Bucket Name", Destination"BucketName")
Ok for the creation form, but the "*" should be present only on the creation stage, not on the edition. So please remove the "*" from the "configuration" tab once the rule has been created.

> 4.         Add an icon next to the rule type for instance <i className="fas fa-coins" /> (see in core-ui https://scality.github.io/core-ui/?path=/story/style-icons--page)
The icon should be on the left (not on the right) - 8px of space between icon & label.

> 5.         "Rule Name" instead of "Name"
Ok

> 6.         States should be with the first letter in Uppercase (Active / Inactive), same in the table.
Ok

> 7.         Workflow type should be with the first letter in Uppercase in the table (here "Replication" instead of "replication")
Ok

> 8.         Add the tooltips next to the label that needs explanation (+ Terry to provide the content)
Terry to wait - if it's the only missing point, we could merge and add the tooltips in another PR.

> 9.         24px spacing between buttons "Cancel" and "Create"
Great

src/react/ui-elements/FormLayout.jsx
```diff
@@ -32,10 +32,16 @@ export const Title = styled.div`
     display: flex;
 
     text-transform: capitalize;
-    margin-bottom: 40px;
+    margin-bottom: 24px;
     font-size: 19px;
 `;
```
theocerutti: Just to be sure: I changed the margin-bottom to 24px for ALL titles (<Title/> Component) is that what I needed to do or is it just in this case?

Cuervino: No, just for the case of Replication, which is a specific one.

Cuervino (approved):
Good!
The tooltips will be done in another issue/PR.

theocerutti:
@bert-e approve

src/react/workflow/WorkflowContent.jsx
```diff
@@ -23,6 +24,15 @@ type Props = {
 
 export const InfoWarning = ({ title }: { title: string}) => <Warning iconClass='fas fa-2x fa-info-circle' title={title} />;
 
+const ReplicationTitle = styled(Title)`
```
nicolas2bert: These styled components can be moved to the `../ui-elements//TableKeyValue2` and be renamed `Title` and `Subtitle`. That would help UX consistency.

src/react/workflow/WorkflowList.jsx
```diff
@@ -8,8 +8,13 @@ import { ListSection } from '../ui-elements/ListLayout3';
 import React from 'react';
 import type { Rules } from '../../types/config';
 import { push } from 'connected-react-router';
+import styled from 'styled-components';
 import { useDispatch } from 'react-redux';
 
+const WorkflowTypeCell = styled.span`
```
nicolas2bert: We might want to create a `ui-elements/Utility.js` file and import it from there for reusability.

src/react/workflow/replication/Replication.jsx
```diff
@@ -28,6 +28,10 @@ const ReplicationContainer = styled.div`
     }
 `;
 
+const ReplicationIcon = styled.i`
```
nicolas2bert: Let's avoid having too many styled components that will affect readability. Can we set an CSS rule for `<T.Key>`'s icon in `../../ui-elements/TableKeyValue2` ?

src/react/workflow/replication/Replication.jsx
```diff
@@ -118,7 +125,7 @@ function Replication({ streams, bucketList, locations, workflowDetails, showEdit
                     </T.GroupName>
                     <T.GroupContent>
                         <T.Row>
-                            <T.Key> Name </T.Key>
+                            <T.Key required={createMode}> Rule Name </T.Key>
```
nicolas2bert: Name is not required, we generate one for you if missing. CC @Cuervino

nicolas2bert (commented):
It looks like your tests are not passing, you can fix them running:
- for linting: npm run test:lint
- for unit: npm run test

src/react/workflow/replication/Replication.jsx
Cuervino: Ah, well spotted.

src/react/workflow/WorkflowContent.jsx
theocerutti: done

src/react/workflow/WorkflowList.jsx
theocerutti: done

src/react/workflow/replication/Replication.jsx
theocerutti: done

src/react/workflow/replication/Replication.jsx
theocerutti: done

Cuervino (approved):
LGTM

src/react/ui-elements/FormLayout.jsx
```diff
@@ -36,9 +36,9 @@ import styled from 'styled-components';
 export const Title = styled.div`
     display: flex;
     text-transform: capitalize;
-    color: ${props => props.theme.brand.textPrimary};
     margin-bottom: 40px;
-    font-size: 24px;
+    font-size: 19px;
```
nicolas2bert: should we use the theme's sizes?

Cuervino: yes, it's always better.

nicolas2bert (approved):

PR #110, opened 2021-04-21
ZENKO-3373: Core-UI paddings / font-size / border-radius

theocerutti:


src/react/ui-elements/Container.jsx
```diff
@@ -1,3 +1,4 @@
+import { padding } from '@scality/core-ui/dist/style/theme';
 import styled from 'styled-components';
 
 const MainContainer = styled.div`
```
nicolas2bert, line 4: Should we set the default `font-size`, `color`, ... value here instead of in `style.css`'s body?

src/react/databrowser/objects/ObjectDelete.jsx
```diff
@@ -17,18 +18,18 @@ const Files = styled.div`
...
     overflow-y: scroll;
-    margin: 10px 0px;
+    margin: ${padding.small} 0px;
     border: 1px solid ${props => props.theme.brand.borderLight};
 `;
 
 const Description = styled.div`
-    font-size: 14px;
-    margin-top: 2px;
+    font-size: ${fontSize.base};
```
nicolas2bert: Do we need to still provide default/base font-size?
If we set a default font-size, should this component inherit from it? Same logic can be apply for the following `font-size: 14px;`.

Cuervino (changes requested):
Could you please set the "padding-top" to 16px (base) for:
- the account list
image
- the Data Browsing - bucket list
image
- the Data Browsing - object-list
image

src/react/databrowser/objects/ObjectDelete.jsx
theocerutti: done

src/react/ui-elements/Container.jsx
theocerutti: done

nicolas2bert (approved):

src/react/databrowser/objects/ObjectListTable.jsx
```diff
@@ -12,12 +12,13 @@ import InfiniteLoader from 'react-window-infinite-loader';
...
 import type { ObjectEntity } from '../../../types/s3';
 import { formatBytes } from '../../utils';
+import { padding } from '@scality/core-ui/dist/style/theme';
 import { push } from 'connected-react-router';
 import styled from 'styled-components';
 
 export const Icon = styled.i`
-    margin-right: 5px;
-    margin-left: ${props => props.isMargin ? '15px' : '0px'};
+    margin-right: ${padding.smaller};
```
nicolas2bert, line 20: Why is not the context's theme used?

nicolas2bert (changes requested):

nicolas2bert (approved):

Cuervino (approved):
LGTM!

theocerutti:
@bert-e approve

PR #111, opened 2021-04-21
Edited README per request

wabernat:
@nicolas2bert: I don't know the zenko-ui build system, so I have pushed my suggestions, but not issued a PR. If that's a reasonable next step, let me know, and I'll proceed with it. Otherwise, you can cherry-pick this onto your next PR. Whatever works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,20 +4,20 @@
 
 Zenko UI is our portal to manage offline Zenko instances.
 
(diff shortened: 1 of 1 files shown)


nicolas2bert:
I have taken to liberty of closing your PR since I used your commit here: https://github.com/scality/zenko-ui/pull/113

PR #120, opened 2021-04-23
ZENKO-3409: Remove quota feature in Storage Location Edition

theocerutti:


Cuervino (changes requested):
1. Even if the rule should be "if one or more fields are visible show advanced options", I don't see "Advanced Options" in my platform example.
2. Could you please add the icon (same as in the Workflow ediiton tab)?
image

nicolas2bert (approved):

Cuervino (approved):
Top!

theocerutti:
@bert-e approve

PR #135, opened 2021-04-28
ZENKO-3439 create access key

nicolas2bert:
Screen Shot 2021-04-28 at 4 17 41 PM

### DIFF
--- a/src/react/account/details/Properties.jsx
+++ b/src/react/account/details/Properties.jsx
@@ -1,7 +1,7 @@
 // @flow
-import type { Account } from '../../../../types/account';
-import AccountInfo from './AccountInfo';
-import AccountKeys from './AccountKeys';
+import type { Account } from '../../../types/account';
+import AccountInfo from './properties/AccountInfo';
+import AccountKeys from './properties//AccountKeys';
 import { AutoSizer } from 'react-virtualized';
 import React from 'react';
 import styled from 'styled-components';
@@ -17,7 +17,7 @@ const Container = styled.div`
     width: ${props => props.width}px;
 `;
 
-function AccountProperties({ account }: Props) {
+function Properties({ account }: Props) {
     return (
         <AutoSizer>
             {({ height, width }) => (
@@ -30,4 +30,4 @@ function AccountProperties({ account }: Props) {
     );
 }
 
-export default AccountProperties;
+export default Properties;
--- a/src/js/mock/IAMClient.js
+++ b/src/js/mock/IAMClient.js
@@ -1,7 +1,7 @@
 // @noflow
-import type { AccessKey } from '../../types/user';
+import type { IamAccessKey } from '../../types/user';
 
-export const accountAccessKeys: Array<AccessKey> = [
+export const accountAccessKeys: Array<IamAccessKey> = [
     {
         AccessKeyId: 'LEAST_RECENT_KEY_BBB',
         Status: 'Active',
--- a/src/js/mock/managementClient.js
+++ b/src/js/mock/managementClient.js
@@ -1,13 +1,14 @@
 // @flow
 
+import type { AccessKey, Account, SecretKey } from '../../types/account';
 import type {
+    ApiAccountKeyResponse,
     ApiAccountResponse,
     ApiConfigurationResponse,
     ManagementClient as ManagementClientInterface,
 } from '../../types/managementClient';
 import type { ConfigurationOverlay, Location, Replication } from '../../types/config';
 import type { APIWorkflows } from '../../types/workflow';
-import type { Account } from '../../types/account';
 import { ApiErrorObject } from './error';
 import { toLocationType } from '../../types/config';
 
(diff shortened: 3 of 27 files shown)


nicolas2bert:
"Success message" reviewed by @wabernatScality.

src/react/account/details/SecretKeyModal.jsx
```diff
@@ -0,0 +1,73 @@
...
+        <Modal
+            id="secret-key-modal"
+            close={handleClose}
+            footer={
+                <div>
+                    <Button id='object-delete-cancel-button' outlined onClick={handleClose} size='small' text='Close'/>
+                </div>
+            }
+            isOpen={true}
+            title='Create access key'>
```
terrydervaux:
```suggestion
            title='Create root user access keys'>
```

src/react/account/details/SecretKeyModal.jsx
```diff
@@ -0,0 +1,73 @@
...
+            <ModalBody>
+                <Banner
+                    title="Success"
+                    icon={<i className="fas fa-check" />}
+                    variant="success"
+                >
+                    A root user access key ID and its secret access key have been created. <br/>
+                    You cannot retrieve this again, so copy it carefully and securely now. <br/>
+                    If you lose this access key, you will have to create a new one. You can do this at any time. <br/>
```
terrydervaux:
```suggestion
                    If you lose this secret access key, you will have to create a new one. You can do this at any time. <br/>
```

terrydervaux (commented):
Please, add the first modal describe in the Jira ticket.

src/react/account/details/Properties.jsx
```diff
@@ -36,12 +37,19 @@ function Properties({ account }: Props) {
...
     // TODO: Should we let the user delete accounts that still owns buckets.
     return (
         <TableContainer>
             <DeleteConfirmation show={showDelete} cancel={handleDeleteCancel} approve={handleDeleteApprove} titleText={`Are you sure you want to delete account: ${account.userName} ?`}/>
+            <SecretKeyModal account={account} />
             <ButtonContainer>
-                <Button icon={<i className="fas fa-trash" />} onClick={handleDeleteClick} size="default" variant="danger" text='Delete Account' />
+                <Button id='delete-account-btn' icon={<i className="fas fa-trash" />} onClick={handleDeleteClick} size="default" variant="danger" text='Delete Account' />
+                {/* TODO: Create access key button  might be displayed somewhere else. */}
+                <Button icon={<i className="fas fa-plus" />} size="default" onClick={handleAccessKeyCreate} text='Create access key'/>
```
jonathan-gramain: maybe capitalize the tooltip: "Create Access Key" (only if consistent with the other tooltips around)

src/react/account/details/SecretKeyModal.jsx
jonathan-gramain: same comment about capitalizing

src/react/account/details/SecretKeyModal.jsx
```diff
@@ -0,0 +1,73 @@
...
+            isOpen={true}
+            title='Create access key'>
+            <ModalBody>
+                <Banner
+                    title="Success"
+                    icon={<i className="fas fa-check" />}
+                    variant="success"
+                >
+                    A root user access key ID and its secret access key have been created. <br/>
+                    You cannot retrieve this again, so copy it carefully and securely now. <br/>
```
jonathan-gramain: I'd rather use the impersonal form "This cannot be retrieved again, so copy... a new one will have to be created. It can be done at any time.", but @wabernatScality can chime in!

src/react/account/details/SecretKeyModal.jsx
```diff
@@ -0,0 +1,73 @@
...
+                    If you lose this access key, you will have to create a new one. You can do this at any time. <br/>
+                </Banner>
+                <Table id='secret-key-table' style={{ marginTop: '16px' }}>
+                    <T.Body>
+                        <T.Row>
+                            <T.Key> Account name </T.Key>
```
jonathan-gramain: ditto regarding capitalization

jonathan-gramain (approved):

src/react/account/details/Properties.jsx
nicolas2bert: We have chosen: "Root user" "Access key ID" and "Secret Access key" :) Thank you.

ilkescality (approved):
LGTM!

Cuervino (changes requested):
image

image

image

Cuervino (changes requested):
image
1. "Root user Access keys"
2. "Root user Access keys"
3. "root Access keys"
4. "Access key ID"
5. "Create Access key"
6. Tootlip too far from the string (and too close to the creation date). Set it closer to the Access key ID string (especially that we know the length of the string)
7. Button too small (27 px instead of 32px high), icon size too small, and the style doesn't seem to be the variant "Delete"
8. Delete button too far on the right, better to make it closer to the other columns

It will prevent having an issue of tooltip under the right column.
image

image
9. The font size is too small (it's 12px, should be 14px like other buttons)
10. "Secret Access key"

image
11. it would be nice that this action would be located at the same place, not dependent on the size of the string

image
12. the button "Delete Account" also looks like not being in the variant "Delete"

nicolas2bert:
1 2 3 4 5 6 7 done
8: we follow what already existed for button "actions" (edit, delete) in locations' list. We might want to change it everywhere. To be discuss :)
9 10 done
11 Is it ok if we work on this comment later?
12 done

Cuervino (approved):
> 8: we follow what already existed for button "actions" (edit, delete) in locations' list. We might want to change it everywhere. To be discuss :)
You're right, so let's keep it like that!

> 11 Is it ok if we work on this comment later?
Sure!

Nicely done !

nicolas2bert:
@bert-e approve

PR #139, opened 2021-04-30
ZENKO-3469 Test object Properties page

nicolas2bert:


JBWatenbergScality:
Duplicate with https://github.com/scality/zenko-ui/blob/development/1.1/src/react/databrowser/objects/details/__tests__/Properties.test.jsx

PR #140, opened 2021-05-03
ZENKO-347 handle IAM errors

nicolas2bert:
Now that AWS IAM calls (create/list/delete access key) are being used, we should return clearer error messages.

### DIFF
--- a/src/types/aws.js
+++ b/src/types/aws.js
@@ -1,0 +1,10 @@
+// @flow
+
+export type AWSError = {
+  +code: string, // a unique short code representing the error that was emitted.
+  +message: string, // a longer human readable error message
+  +retryable?: boolean, // whether the error message is retryable.
+  +statusCode?: number, // in the case of a request that reached the service, this value contains the response status code.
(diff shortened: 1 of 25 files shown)


src/types/aws.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+
+export type AWSError = {
+  +code: string, // a unique short code representing the error that was emitted.
+  +message: string, // a longer human readable error message
+  +retryable?: boolean, //whether the error message is retryable.
```
anurag4DSB: nit:
```suggestion
  +retryable?: boolean, // whether the error message is retryable.
```
Also, what do you mean when you say `error message is retryable`? Do you mean the request?

nicolas2bert: yes, retrying request when error occurred. 
I just have copied-pasted descriptions from the Property Details sections. https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Response.html

anurag4DSB (approved):

ilkescality (approved):

nicolas2bert:
@bert-e approve

PR #144, opened 2021-05-06
ZENKO-3413: adjust date format for tables

theocerutti:
Date format updated from 
image
to
image

(only for tables)

Tables updated:
* AccountKeys / Created On
* Bucket Browsing / Created On

### DIFF
--- a/src/react/databrowser/buckets/BucketList.jsx
+++ b/src/react/databrowser/buckets/BucketList.jsx
@@ -8,6 +8,7 @@ import { useFilters, useSortBy, useTable } from 'react-table';
 import { AutoSizer } from 'react-virtualized';
 import { FixedSizeList } from 'react-window';
 import type { S3BucketList } from '../../../types/s3';
+import { TextAligner } from '../../ui-elements/Utility';
 import { formatShortDate } from '../../utils';
 import { getLocationTypeFromName } from '../../utils/storageOptions';
 import { push } from 'connected-react-router';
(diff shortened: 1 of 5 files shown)


nicolas2bert (commented):
What about object page's **Modified on** column?

theocerutti:
> What about object page's Modified on column?

It already has the good format
image

nicolas2bert (approved):

Cuervino (changes requested):
Looks good!

It was not in the requirement, but could you please make those 2 columns aligned to the right (as they are numerical)?
From the visual point of view, it would mainly change the header position, the content would not move.
image

src/react/databrowser/buckets/BucketList.jsx
```diff
@@ -46,7 +46,12 @@ export default function BucketList({ selectedBucketName, buckets, locations }: P
         {
             Header: 'Created on',
             accessor: 'CreationDate',
-            Cell: ({ value }) => { return formatShortDate(new Date(value));},
+            headerStyle: { textAlign: 'right', paddingRight: '32px' },
+            Cell({ value }: { value: string }) { return (
+                <div style={{ textAlign: 'right', paddingRight: '16px' }}>
```
nicolas2bert: Can you also use the TextAligner component here?

Cuervino (approved):
LGTM!

nicolas2bert (approved):

theocerutti:
@bert-e approve

PR #153, opened 2021-05-26
ZENKO-3539: improve data browsing ux

theocerutti:
image


nicolas2bert:
Why "Version ID" and "Modified on" items are not aligned the same way? @Cuervino

Cuervino:
> Why "Version ID" and "Modified on" items are not aligned the same way? @Cuervino

Version ID is a string of character that is not numerical, but I consider a date as a numerical value. 
Aligning on the right makes it easier to compare values. It especially makes sense for the Size column, but IMO the date could also follow this rule, where the truncated content for Version ID is more a "name" / non-numerical value.

nicolas2bert:
For date, I do not see why aligning left would make comparing dates harder. I assumed each dates were all the same size as well as number characters.

Cuervino:
You're right, for the date format in table (especially in the full numerical format, with leading zeros), it doesn't matter that much, there is no definitive consensus about that (right or left alignment) among the Usability community.

Cuervino (changes requested):
1. ZENKO-3538: Reduce the height of the Metadata Search field
LGTM

2. ZENKO-3540: Amend the Version ID column truncation logic
a. In my test, a space is displayed between the [...] and the right part of truncation, it should be removed.
image

b. On hover, the tooltip should not be hidden (and should display the full string, as in the Summary tab version)
image

3. ZENKO-3537: Align the header checkbox vertically in the Table
LGTM

4. ZENKO-3536: Data Browser: Columns Size and Modified align to the right
a. Ok for the Size column

b. On default zoom level, the header and the content are not aligned.
image

c. Moreover, when zooming, the alignement brokes even more, and the right side of the content disappears.
image

nicolas2bert:
I will push for aligning left by default and when necessary (better readability...) aligning on the right. 
For this table, visually, dates seem to be floating around.

Cuervino:
Even if there is no final consensus, IMO, the readability is better when we align dates on the right.
For instance, on Excel or on Google Spreadsheet, when the user types a data that is identified as a text, the cell stays left-aligned, but when the data is identified as a date, the alignement automatically switches to right, because this is a better legibility.

Cuervino (commented):
Everything looks good to me.

One bug happened: sometimes, the truncation on the versionID in the tab is lost. Is it related with this PR?

image

Cuervino (approved):
LGTM!

src/react/databrowser/objects/ObjectListTable.jsx
```diff
@@ -134,23 +152,36 @@ export default function ObjectListTable({ objects, bucketName, toggled, isVersio
         initialState: { hiddenColumns },
     }, useFilters, useSortBy, useFlexLayout);
 
+    const refList = useCallback((ref) => {
```
nicolas2bert, line 157: Here we are assuming that scrollbar will be 8px wide and so that it will move our cells 8px to the left.
That seems to be true in Chrome but Firefox scrollbar is not taking any width and so we end up with unaligned header and content. 

Screen Shot 2021-07-01 at 10 47 28 AM

src/react/ui-elements/MiddleEllipsis.jsx
```diff
@@ -0,0 +1,117 @@
+//@flow
```
nicolas2bert, line 1: Cool component, should we move it to core-ui? @Cuervino 
If yes, we could do it in a following PR @theocerutti.

src/react/databrowser/objects/ObjectListTable.jsx
theocerutti: It seems it is only on MacOS.
I also noticed that on my side firefox and google chrome scrollbar size has a tiny difference. (that affects alignment on Firefox)

image

src/react/ui-elements/MiddleEllipsis.jsx
theocerutti: ~~Just one thing : I added "width" props in MiddleEllipsis for a workaround.
Indeed, MiddleEllipsis doesn't work fine in an AutoSizer of react-virtualized.
That's why MiddleEllipsis component needs to rerender if AutoSizer sizes has changed.~~

It seems the workaround is no longer needed.

nicolas2bert (approved):

nicolas2bert:
Please don't forget to squash before merging. Thank you :)

theocerutti:
I re-add the workaround for MiddleEllipsis because the bug still occurs.



Step of reproduction:
It seems the bug occurs only if the MiddleEllipsis is in an AutoSizer AND in a long list (it is weird but the bug only happens when the list is large enough to display a scrollbar).

1/ You have to resize the window with the "resize" button of the window.
image

2/ Go fullscreen

Then the MiddleEllipsis displays as:
image


Also, I noticed that the MiddleEllipsis is displaying at full length (without ellipsis) on first render.
I applied an "opacity: 0" only if it has not yet calculated the ellipsis.

theocerutti:
@bert-e approve

PR #158, opened 2021-06-08
ARTESCA-1355: reduce list of storage location

theocerutti:
image


Cuervino (approved):
LGTM!

src/react/utils/storageOptions.js
```diff
@@ -48,11 +49,14 @@ export function getLocationBucketName(locationConstraint, locations) {
 export function selectStorageOptions(
     capabilities: $PropertyType<InstanceStateSnapshot, 'capabilities'>,
     labelFn?: LabelFunction,
+    exceptOldVersion: boolean = true,
 ): Array<StorageOptionSelect> {
     return Object.keys(storageOptions).filter(o => {
-        const isOldVersion = !!storageOptions[o].isOldVersion;
-        if (isOldVersion) {
-            return;
+        if (exceptOldVersion) {
```
nicolas2bert: Let's return a boolean from the filter callback.

src/react/backend/location/LocationDetails/storageOptions.js
```diff
@@ -62,37 +60,49 @@ export const storageOptions: ({ [name: string]: any }) = {
...
+        formDetails: LocationDetailsAwsCustom,
         supportsVersioning: true,
-        supportsReplicationTarget: false,
+        supportsReplicationTarget: true,
         supportsReplicationSource: true,
         hasIcon: false,
-        checkCapability: 'locationTypeHyperdriveV2',
+        checkCapability: 'locationTypeCephRadosGW',
+        ingestCapability: 'cephIngestLocation',
+        isOldVersion: true,
```
nicolas2bert: Let's call it `hidden` instead of `isOldVersion`.

nicolas2bert (commented):
@Cuervino  @terrydervaux What should we do if the user has already created a "now hidden" location (e.g. NFS Mount location)? Should user be able to see it / edit it / delete it? 
@theocerutti "legacy support" tests would be needed.

src/react/utils/storageOptions.js
Cuervino: > 
> 
> @Cuervino @terrydervaux What should we do if the user has already created a "now hidden" location (e.g. NFS Mount location)? Should user be able to see it / edit it / delete it?
> @theocerutti "legacy support" tests would be needed.

This case will not happen outside of our test environment, no?

src/react/account/details/Locations.jsx
```diff
@@ -122,7 +122,7 @@ function Locations() {
     }, useFilters, useSortBy);
 
     // NOTE: empty state component
-    if (locations.length === 0) {
+    if (Object.keys(locations).length === 0) {
```
nicolas2bert, line 125: Why do we need to use `Object.keys` method?

theocerutti: Because `locations` is an object and not an array. (typeof Locations)
```
export type Locations = $ReadOnly<PerLocationMap<Location>>;
export type PerLocationMap<T> = {
    [LocationName]: T,
};
```

```
console.log([].length) // 0
console.log({}.length) // undefined
```

nicolas2bert: Good catch, thank you :)

src/react/utils/storageOptions.js
```diff
@@ -48,11 +49,14 @@ export function getLocationBucketName(locationConstraint, locations) {
...
-        const isOldVersion = !!storageOptions[o].isOldVersion;
-        if (isOldVersion) {
-            return;
+        if (exceptHidden) {
+            const hidden = !!storageOptions[o].hidden;
+            if (hidden) {
+                return false;
+            }
         }
         return o;
```
nicolas2bert: `return true` instead of `return o`?
From the `.filter()` doc: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter 
```
callbackFn
Function is a predicate, to test each element of the array. Return a value that coerces to true to keep the element, or to false otherwise.
```

theocerutti: yes you're right, I will modify it right now

nicolas2bert (approved):
You can squash and merge :)

theocerutti:
@bert-e approve

PR #164, opened 2021-07-07
ZENKO-3301: Lifecycle Expiration

theocerutti:
- [ ] 1. Decouple rule type from Replication component - we might need to introduce a new “Workflow” component + tests.
(https://github.com/scality/zenko-ui/pull/165)

- [ ] 2. Create/Edit expiration workflow Redux logic + tests.

- [ ] 3. Expiration workflow “source” part logic + tests.

- [ ] 4. Expiration workflow creation/edition layout + tests.

- [ ] 5. List expiration workflow Redux logic + new row type + tests.

- [ ] 6. Delete expiration + confirmation message + tests.


src/react/workflow/workflow/Workflow.jsx
```diff
@@ -0,0 +1,132 @@
+//@flow
```
nicolas2bert: Let's move `src/react/workflow/workflow/Workflow.jsx` to `src/react/workflow/WorkflowEditor.jsx`

src/react/workflow/replication/Replication.jsx
```diff
@@ -1,85 +1,53 @@
 // @flow
```
nicolas2bert: Let's move `src/react/workflow/replication/Replication.jsx` to `src/react/workflow/workflowTypes/Replication.jsx`.
We will have soon `src/react/workflow/workflowTypes/Expiration.jsx` for the expiration form.

src/react/workflow/details/Configuration.jsx
```diff
@@ -62,7 +54,7 @@ function Configuration({ wfSelected, replications, bucketList, locations, showEd
                         </T.BannerContainer>
                         <Button icon={<i className="fas fa-trash" />} text="Delete Workflow" variant='buttonDelete' onClick={handleOpenDeleteModal} />
```
nicolas2bert: "Banner" and "deletion button" could be moved to `src/react/workflow/WorkflowEditor.jsx` and only be shown when createMode is set to `false`. That way we might avoid having to (re)define `showEditWorkflowNotification` in three deferent component, hence avoiding extra rendering.

nicolas2bert (commented):
First iteration on commit: 9369d2da2ce6d6e12db1cb3cc038e659ad1594bd

src/react/workflow/details/Configuration.jsx
elenamarcova: Code review is expected from @nicolas2bert .

PR #168, opened 2021-08-18
Update Core-UI, use V2 component, use REM unit, fix broken UI

theocerutti:


Cuervino (changes requested):
**Functional review**

- The modal seems not wide enough on the Root user Keys modal
image

- The breadcrumb is not aligned
image

- The Metadata tab and the Tags tab are kind of broken (in the Data Browser, for Objects)
image


**Minor issues:**
- Still the issue of columns alignement in tables, I'm not sure it's covered by this PR.
image

- In the tables, cursor should not always be "action"
image

**Otherwise,** 
I checked the buttons, the zoom, the rem, everything looks rather neat :-)

Cuervino (changes requested):
1. Now, a lot of columns in tables are not aligned 
For instance, this column should be aligned on the right.
image

2. The issue with the columns looks like creating a horizontal slider for most of the container with tables.
Remove the slider:
image

Cuervino (approved):
LGTM!

nicolas2bert (approved):

theocerutti:
@bert-e approve

PR #172, opened 2021-09-30
feature: ZKUI-7 Add new location for Artesca

miniscruff:


mrepop (approved):

src/react/backend/location/__tests__/LocationEditor.test.jsx
```diff
@@ -12,12 +12,13 @@ describe('LocationEditor', () => {
...
         expect(options.at(1).find('span').text()).toBe('Scality RING with S3 Connector');
-        expect(options.at(2).find('span').text()).toBe('Amazon S3');
-        expect(options.at(3).find('span').text()).toBe('Google Cloud Storage');
-        expect(options.at(4).find('span').text()).toBe('Microsoft Azure Blob Storage');
-        expect(options.at(5).find('span').text()).toBe('Local Filesystem');
+        expect(options.at(2).find('span').text()).toBe('Scality Artesca S3');
+        expect(options.at(3).find('span').text()).toBe('Amazon S3');
+        expect(options.at(4).find('span').text()).toBe('Google Cloud Storage');
+        expect(options.at(5).find('span').text()).toBe('Microsoft Azure Blob Storage');
+        expect(options.at(6).find('span').text()).toBe('Local Filesystem');
```
naren-rajendran, line 22: Not a fan of index based tests, could this be changed? and they are not in alphabetical order, is there a reason why? maybe we want to list scality types first?

miniscruff: I agree on the indexing but I felt it would be a bit much to restructure, but it did seem this order was important as it matched the other projects. Could not tell you if that is true though, I just added Artesca after Ring as it made sense to me.

src/react/backend/location/LocationDetails/storageOptions.js
```diff
@@ -32,6 +32,17 @@ export const storageOptions: ({ [name: string]: any }) = {
...
+    'location-scality-artesca-s3-v1': {
+        name: 'Scality Artesca S3',
+        short: 'ARTESCA',
+        formDetails: LocationDetailsAwsCustom,
+        supportsVersioning: true,
+        supportsReplicationTarget: true,
+        supportsReplicationSource: true,
+        hasIcon: false,
+        checkCapability: 'locationTypeS3Custom',
+        ingestCapability: 's3cIngestLocation',
```
nicolas2bert: I will remove this property since we do not support ingestion from Artesca S3 location type.

src/react/backend/location/__tests__/LocationEditor.test.jsx
nicolas2bert: From what I remember, the order matters.

naren-rajendran: @nicolas2bert scality types first?

nicolas2bert (approved):

src/react/backend/location/__tests__/LocationEditor.test.jsx
nicolas2bert: I would assume so, @Cuervino or @thomasdanan might want to confirm.

miniscruff:
gonna clean up the commits and then approve

src/react/backend/location/__tests__/LocationEditor.test.jsx
miniscruff: Checked with Thomas that Artesca should be before Ring, changing now.

PR #178, opened 2021-10-14
ZKUI-11 release pipeline hash retrieval is not reliable

nicolas2bert:


miniscruff (approved):
we could just use `--short=12` but fair point

tcarmet (approved):

nicolas2bert:
@bert-e approve

PR #181, opened 2021-10-21
Add object lock and versioning setting in create bucket page

ChengYanJin:
Add object lock and versioning setting in create bucket page


ChengYanJin:
Wrong branch name, will re-create a pull request...

PR #182, opened 2021-10-21
ARTESCA-2233: add object lock and versioning setting in bucket creation

ChengYanJin:
Add object lock and versioning setting in bucket creation
Add prettier configuration


Cuervino (commented):
image

1. Bucket Name > it looks like some input constraint are missing (any Character can be typed?) (It might be out of the ticket).

2. The scrolling area is not correct, the title "Create a new Bucket" and the bottom buttons should be always visible.

image

3. When the Versioning is automatically enabled, it should be in a "Active/disabled mode" (here, it's active, but not "actionable"). 
It looks like you saw that ("disable state of the toggle")

4. "Automatically activated when Object-lock is Active" > text should be SmallerText, with TextSecondary Color

5. Description of the features > text should be TextSecondary Color

image

6. Nice-to-have > I would prefer a "shadow effect" on the non-yet-activable options, rather than visible/invisible, with the Ui-action component disabled (radio button, dropdowns).
It looks like you saw that ("add opacity for the retention setting")

7. In general, I feel like there is a little bit to much explanation text, some could be added in the tooltips. WDYT?

8. Alignement issues
	-Radio button and label
	-"period" and "unit of time"
	
9. I would put the default "unit of time" on "days", not "years". And I would sort it "Days" > "Years"

src/react/databrowser/buckets/BucketCreate.jsx
```diff
@@ -156,15 +203,215 @@ function BucketCreate() {
...
+            objects and the bucket.
+            <br />
+            Enabling Object-lock automatically activates Versioning for the
+            bucket, and you won’t be able to suspend Versioning.
+          </SmallerText>
+        </SpacedBox>
+        {isObjectLockEnabled ? (
+          <>
+            <F.Fieldset direction={'row'}>
+              <F.Label style={{ width: '25%' }}>Default Retention</F.Label>
```
nicolas2bert: Can we remove `style` prop and make a new component?

src/react/databrowser/buckets/BucketCreate.jsx
```diff
@@ -156,15 +203,215 @@ function BucketCreate() {
...
+            you&apos;ll be able to disable the retention itself on edition.
+            <br />
+            Once the bucket is created, you might be blocked from deleting the
+            objects and the bucket.
+            <br />
+            Enabling Object-lock automatically activates Versioning for the
+            bucket, and you won’t be able to suspend Versioning.
+          </SmallerText>
+        </SpacedBox>
+        {isObjectLockEnabled ? (
```
nicolas2bert: ~~`BucketCreate.jsx`is starting to be "a little too" big to read. Can we move the object lock options to its own redux container?~~ Addressed already :)

src/react/databrowser/buckets/ObjectLockRetentionSettings.jsx
```diff
@@ -0,0 +1,177 @@
...
+                    <F.Select.Option value={'YEARS'}>Years</F.Select.Option>
+                    <F.Select.Option value={'DAYS'}>Days</F.Select.Option>
+                  </F.Select>
+                </>
+              );
+            }}
+          />
+        </SpacedBox>
+
+        <SpacedBox style={{ width: '25%' }} ml={8}>
```
nicolas2bert: Duplicated?

ChengYanJin: Actually, this is to define the width of the DAYS/YEARS Select component.

JBWatenbergScality (approved):

src/react/databrowser/buckets/BucketCreate.jsx
ChengYanJin: Addressed here https://github.com/scality/zenko-ui/pull/182/commits/b3b3057752948ff869bd550ff8db3f59da35ea77

nicolas2bert:
Really hard to review, can we have "prettier changes" in a separate PR since it has nothing to do with this one? Also can you clean up/squash some commits?

ChengYanJin:
> Really hard to review, can we have "prettier changes" in a separate PR since it has nothing to do with this one? Also can you clean up/squash some commits?

Hello @nicolas2bert ,

Apologize for the unclean commits!! I just clean up the history.
Thanks for your review ;)

nicolas2bert (approved):
Much cleaner, thank you!

src/react/databrowser/buckets/ObjectLockRetentionSettings.jsx
```diff
@@ -0,0 +1,172 @@
...
+import { spacing } from '@scality/core-ui/dist/style/theme';
+import * as F from '../../ui-elements/FormLayout';
+import React from 'react';
+
+type ObjectLockRetentionSettingsProps = {|
+  +isObjectLockEnabled: boolean,
+|};
+
+export default function ObjectLockRetentionSettings({
+  isObjectLockEnabled,
```
JBWatenbergScality: I think this component don't need that prop. I removed it before but I think my change has been lost during the rebase :(.
If you prefer we can handle it in my PR tomorrow or you can compare the file with the one of my branch.

ChengYanJin: Removed this isObjectLockEnabled prop.

ChengYanJin:
Without ZKUI ticket number in the ticket
Created another pull request: https://github.com/scality/zenko-ui/pull/191

PR #183, opened 2021-10-21
ZKUI-13 switch to github actions

tcarmet:


.github/workflows/release.yaml
```diff
@@ -0,0 +1,43 @@
...
+        uses: docker/build-push-action@v2.7.0
+        with:
+          context: .
+          push: true
+          tags: "registry.scality.com/zenko-ui/zenko-ui:${{ github.event.inputs.tag }}"
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+
+      - name: Create Release
+        uses: softprops/action-gh-release@v0.1.13
```
miniscruff, line 49: this was updated in pensieve like an hour ago, I think it is v1 now

.github/dependabot.yaml
```diff
@@ -0,0 +1,7 @@
+---
```
miniscruff, line 1: while you are here you might want to add the javascript package

.github/workflows/release.yaml
tcarmet: the maintainer of this action is tagging v1 as 0.1.13. I would argue it's rather a v0, but ok. Anyway I prefer to leave it this way.

.github/dependabot.yaml
tcarmet: done

.github/workflows/release.yaml
miniscruff: You sure? I am pretty sure dependabot will open a PR next tick

miniscruff (approved):

nicolas2bert (approved):

.github/workflows/release.yaml
```diff
@@ -0,0 +1,55 @@
...
+        with:
+          node-version: '10'
+          cache: 'npm'
+
+      - name: install dependencies
+        run: npm config set unsafe-perm true && npm ci
+
+      - name: build assets
+        run: npm run build
+
```
nicolas2bert, line 38: are the 2 line breaks on purpose?

tcarmet: no they're not, let me fix that.

PR #190, opened 2021-10-27
bucket-list: Add default retention information in bucket properties panel

JBWatenbergScality:


MonPote (approved):

src/js/mock/S3Client.js
```diff
@@ -105,6 +105,60 @@ export const bucketInfoResponse: BucketInfo = {
   isVersioning: false,
   public: false,
   locationConstraint: '',
+  objectLockConfiguration: {
+    ObjectLockEnabled: 'Disabled',
```
ChengYanJin, line 109: Maybe we should extract 'Enabled' and 'Disabled' 'Suspended' as a property.. 
but that's really minor..

ChengYanJin (approved):

PR #203, opened 2021-11-23
feature: ZKUI-37 support ARWWI

ghost:
This PR introduces the changes needed to make ZKUI work with Vault support for ARWWI.
Linked PR in pensieve-api: https://github.com/scality/pensieve-api/pull/143

https://github.com/scality/zenko-ui/pull/203/commits/1138d1561a00db168e44312d329b10d97b027c63: sets the default role as Storage Manager as Vault was updated according to the recent changes in the design.

### DIFF
--- a/src/js/IAMClient.js
+++ b/src/js/IAMClient.js
@@ -5,6 +5,9 @@ import type { Credentials } from '../types/zenko';
 import IAM from 'aws-sdk/clients/iam';
 import type { IAMClient as IAMClientInterface } from '../types/iam';
 import { getClients } from '../react/utils/actions';
+import { formatRoleSessionName } from '../react/utils';
+
+export const roleName = 'scality-internal/storage-account-owner-role';
 
 export function getAssumeRoleWithWebIdentityIAM(
   state: AppState,
@@ -18,8 +21,8 @@ export function getAssumeRoleWithWebIdentityIAM(
   if (!account || !oidc || !oidc.user) return Promise.reject();
   const assumeRoleParams = {
     idToken: oidc.user.id_token,
-    RoleSessionName: 'app1',
-    roleArn: `arn:aws:iam::${account.id}:role/roleForB`,
+    roleArn: `arn:aws:iam::${account.id}:role/${roleName}`,
+    RoleSessionName: formatRoleSessionName(oidc),
   };
   return stsClient.assumeRoleWithWebIdentity(assumeRoleParams).then(creds => {
     const params = {
--- a/src/js/IAMClient.js
+++ b/src/js/IAMClient.js
@@ -7,7 +7,7 @@ import type { IAMClient as IAMClientInterface } from '../types/iam';
 import { getClients } from '../react/utils/actions';
 import { formatRoleSessionName } from '../react/utils';
 
-export const roleName = 'scality-internal/storage-account-owner-role';
+export const roleName = 'scality-internal/storage-manager-role';
 
 export function getAssumeRoleWithWebIdentityIAM(
   state: AppState,
--- a/src/js/IAMClient.js
+++ b/src/js/IAMClient.js
@@ -1,12 +1,23 @@
 // @noflow
 
-import type { AppState } from '../types/state';
+import type { AppState, AssumeRoleParams, OIDCState } from '../types/state';
 import type { Credentials } from '../types/zenko';
 import IAM from 'aws-sdk/clients/iam';
 import type { IAMClient as IAMClientInterface } from '../types/iam';
 import { getClients } from '../react/utils/actions';
 
-export const roleName = 'scality-internal/storage-manager-role';
+export const rolePathName = 'scality-internal/storage-manager-role';
+
+export function getAssumeRoleWithWebIdentityParams(
+  oidc: OIDCState,
+  accountID: string,
+): AssumeRoleParams {
+  return {
+    idToken: oidc.user.id_token,
+    roleArn: `arn:aws:iam::${accountID}:role/${rolePathName}`,
+    RoleSessionName: oidc.user.profile.sub,
+  };
+}
 
 export function getAssumeRoleWithWebIdentityIAM(
   state: AppState,
@@ -18,11 +29,7 @@ export function getAssumeRoleWithWebIdentityIAM(
   const account = accounts.find(a => a.userName === accountName);
 
   if (!account || !oidc || !oidc.user) return Promise.reject();
-  const assumeRoleParams = {
-    idToken: oidc.user.id_token,
-    roleArn: `arn:aws:iam::${account.id}:role/${roleName}`,
-    RoleSessionName: oidc.user.profile.sub,
-  };
+  const assumeRoleParams = getAssumeRoleWithWebIdentityParams(oidc, account.id);
   return stsClient.assumeRoleWithWebIdentity(assumeRoleParams).then(creds => {
     const params = {
       accessKey: creds.Credentials.AccessKeyId,
--- a/src/js/IAMClient.js
+++ b/src/js/IAMClient.js
@@ -5,7 +5,6 @@ import type { Credentials } from '../types/zenko';
 import IAM from 'aws-sdk/clients/iam';
 import type { IAMClient as IAMClientInterface } from '../types/iam';
 import { getClients } from '../react/utils/actions';
-import { formatRoleSessionName } from '../react/utils';
 
 export const roleName = 'scality-internal/storage-manager-role';
 
@@ -22,7 +21,7 @@ export function getAssumeRoleWithWebIdentityIAM(
   const assumeRoleParams = {
     idToken: oidc.user.id_token,
     roleArn: `arn:aws:iam::${account.id}:role/${roleName}`,
-    RoleSessionName: formatRoleSessionName(oidc),
+    RoleSessionName: oidc.user.profile.sub,
   };
   return stsClient.assumeRoleWithWebIdentity(assumeRoleParams).then(creds => {
     const params = {
--- a/src/react/actions/replication.js
+++ b/src/react/actions/replication.js
@@ -10,7 +10,7 @@ import { networkEnd, networkStart } from './network';
 import type { Replication } from '../../types/config';
 import type { ThunkStatePromisedAction } from '../../types/actions';
 import { push } from 'connected-react-router';
-import { roleName } from '../../js/IAMClient';
+import { rolePathName } from '../../js/IAMClient';
 
 // TODO: Add delete approval process
 export function deleteReplication(
@@ -26,7 +26,7 @@ export function deleteReplication(
       instanceId,
       accountId,
       workflowId: replication.streamId,
-      roleName,
+      rolePathName,
     };
     return managementClient
       .deleteBucketWorkflowReplication(params)
@@ -57,7 +57,7 @@ export function saveReplication(
       workflow: replication,
       bucketName: replication.source.bucketName,
       accountId,
-      roleName,
+      rolePathName,
     };
     const op = replication.streamId
       ? managementClient.updateBucketWorkflowReplication({
--- a/src/react/actions/replication.js
+++ b/src/react/actions/replication.js
@@ -10,6 +10,7 @@ import { networkEnd, networkStart } from './network';
 import type { Replication } from '../../types/config';
 import type { ThunkStatePromisedAction } from '../../types/actions';
 import { push } from 'connected-react-router';
+import { roleName } from '../../js/IAMClient';
 
 // TODO: Add delete approval process
 export function deleteReplication(
@@ -25,6 +26,7 @@ export function deleteReplication(
       instanceId,
       accountId,
       workflowId: replication.streamId,
+      roleName,
     };
     return managementClient
       .deleteBucketWorkflowReplication(params)
@@ -55,6 +57,7 @@ export function saveReplication(
       workflow: replication,
       bucketName: replication.source.bucketName,
       accountId,
+      roleName,
     };
     const op = replication.streamId
       ? managementClient.updateBucketWorkflowReplication({
--- a/src/react/actions/sts.js
+++ b/src/react/actions/sts.js
@@ -7,18 +7,15 @@ import {
 } from './index';
 import type { ThunkStatePromisedAction } from '../../types/actions';
 import { getClients } from '../utils/actions';
+import { getAssumeRoleWithWebIdentityParams } from '../../js/IAMClient';
 
 export function assumeRoleWithWebIdentity(
-  roleArn: string,
+  accountID: string,
 ): ThunkStatePromisedAction {
   return (dispatch, getState) => {
     const { zenkoClient, stsClient } = getClients(getState());
     const { oidc } = getState();
-    const assumeRoleParams = {
-      idToken: oidc.user.id_token,
-      roleArn,
-      RoleSessionName: oidc.user.profile.sub,
-    };
+    const assumeRoleParams = getAssumeRoleWithWebIdentityParams(oidc, accountID);
     return stsClient
       .assumeRoleWithWebIdentity(assumeRoleParams)
       .then(creds => {
--- a/src/react/actions/sts.js
+++ b/src/react/actions/sts.js
@@ -6,7 +6,6 @@ import {
   searchWorkflows,
 } from './index';
 import type { ThunkStatePromisedAction } from '../../types/actions';
-import { formatRoleSessionName } from '../utils';
 import { getClients } from '../utils/actions';
 
 export function assumeRoleWithWebIdentity(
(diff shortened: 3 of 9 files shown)


src/js/IAMClient.js
```diff
@@ -18,8 +20,8 @@ export function getAssumeRoleWithWebIdentityIAM(
   if (!account || !oidc || !oidc.user) return Promise.reject();
   const assumeRoleParams = {
     idToken: oidc.user.id_token,
-    RoleSessionName: 'app1',
-    roleArn: `arn:aws:iam::${account.id}:role/roleForB`,
+    roleArn: `arn:aws:iam::${account.id}:role/${roleName}`,
+    RoleSessionName: (oidc.user) ? oidc.user.profile.email : '',
```
miniscruff:
```suggestion
    RoleSessionName: oidc.user?.profile.email || '';
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

src/react/actions/replication.js
```diff
@@ -25,6 +27,10 @@ export function deleteReplication(
       instanceId,
       accountId,
       workflowId: replication.streamId,
+      role: {
+        roleName: roleName,
+        RoleSessionName: (oidc.user) ? oidc.user.profile.email : '',
```
miniscruff: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

src/react/actions/replication.js
```diff
@@ -47,9 +53,12 @@ export function saveReplication(
 ): ThunkStatePromisedAction {
   return (dispatch, getState) => {
     const state = getState();
+    const { oidc } = getState();
     const { managementClient, instanceId } = getClients(state);
     const accountId = getAccountId(state);
     dispatch(networkStart('Creating replication'));
+    replication.roleName = roleName;
+    replication.RoleSessionName = (oidc.user) ? oidc.user.profile.email : '';
```
miniscruff: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

Tbh at this point a functhion might be better... :smirk:

ghost: Yes, many thanks for the feedback!

ghost: As the roleSessionName is now computed by pensieve-api dynamically (see https://github.com/scality/pensieve-api/pull/143), I removed most of these lines.

I created a new function anyway, as it may be helpful in the future: https://github.com/scality/zenko-ui/pull/203/files#diff-ff376b58fc91513c886b2a8892cd7f7f1303edc5b9be29248b666d2a5d0c8985R43

ghost: We use this function here: https://github.com/scality/zenko-ui/pull/203/files#diff-e728917123f674a0661856cb2a62fe471f3bbdc2435af7063dfdb4b961f7b15aR21
and here https://github.com/scality/zenko-ui/pull/203/files#diff-524d780d05f66cc8206f06f1bb1dd913e4932608ba6d60a6be5cf1f746113ba7R25

ghost:
## Zenko-operator testing with ARWWI

This PR is part of a set of 3 PRs whose objective is to allow Zenko-UI to work correctly following the implementation of AssumeRoleWithWebIdentity in Vault. It is no longer possible to use a random role, and we also want to make the `roleSessionName` dynamic depending on the identity linked to the JWT from keycloak.

Associated PRs:
- https://github.com/scality/Vault/pull/1794
- https://github.com/scality/pensieve-api/pull/143

First, the ARWWI call from the UI is made like so:

```
POST /sts HTTP/1.1
Host: ui.zenko.local
Connection: keep-alive
Content-Length: 1467
X-Amz-Content-Sha256: ***
X-Amz-User-Agent: aws-sdk-js/2.845.0 promise
User-Agent: ***
Accept: */*
Sec-GPC: 1
Origin: http://ui.zenko.local
Referer: http://ui.zenko.local/accounts/aaa
Accept-Encoding: gzip, deflate
Accept-Language: fr-FR,fr;q=0.9

```

Where the Form Data is:

```
Action: AssumeRoleWithWebIdentity
DurationSeconds: 900
RoleArn: arn:aws:iam::248787231990:role/scality-internal/storage-manager-role
RoleSessionName: zenko@example.com
Version: 2011-06-15
WebIdentityToken: <TOKEN_FROM_KEYCLOAK>
```

If the identity has the `StorageManager` role, then ARWWI calls are allowed:

image

image

Otherwide, an error message is displayed:

image



Then, the UI performs some replication workflows calls to pensieve-API. The latter has been updated to use a dynamic roleName (from URL params) and roleSessionName (extracted from the JWT) from the information provided by the UI (referer URL here is `http://ui.zenko.local/api/v1/instance/969ce289-5fa7-473e-9fff-012a218525ea/account/248787231990/workflow/search?roleName=scality-internal%2Fstorage-manager-role`):

image

Vault's logs are then correctly displaying the current session information:

1. When the UI calls ARWWI

```
{"name":"Vault","method":"Router.doRoute","httpMethod":"POST","url":"/sts","requestData":
{"Action":"AssumeRoleWithWebIdentity","DurationSeconds":"900","RoleArn":"arn:aws:iam::248787231990:role/scality-
internal/storage-manager-role","RoleSessionName":"9afe9b78-7e9f-4944-8573-413e7d6ee285","Version":"2011-06-
15"},"time":1637848528697,"req_id":"cf170e4f5d43bf9a213b","level":"info","message":"request received","hostname":"dev-
connector-vault-5d4d9b489f-cbn97","pid":40}
```

2. When Pensieve-API calls ARWWI

```
{"name":"Vault","method":"Router.doRoute","httpMethod":"POST","url":"/","requestData":
{"Action":"AssumeRoleWithWebIdentity","RoleArn":"arn:aws:iam::248787231990:role/scality-internal/storage-manager-
role","RoleSessionName":"9afe9b78-7e9f-4944-8573-413e7d6ee285-pensieve-api","Version":"2011-06-
15"},"time":1637848528621,"req_id":"f6c727789f4388fc3423","level":"info","message":"request received","hostname":"dev-
connector-vault-5d4d9b489f-cbn97","pid":40}
```

src/react/utils/index.js
```diff
@@ -40,6 +40,10 @@ export function formatDate(d) {
   return `${d.toDateString()} ${d.toTimeString().split(' ')[0]}`;
 }
 
+export function formatRoleSessionName(oidc) {
+  return `${oidc.user?.profile.email || 'zenko-ui'}`;
```
JBWatenbergScality: Is fallbacking to `zenko-ui` when the email of the user is not defined something planed in any tech design (I'm not aware of the feature details) ? I'm just asking because if it is not the case IMO when the user profile is missing an email we should likely prompt him with a message saying so, like "Invalid profile, missing required email". In the current implementation an user without email will be prompted with a not authorized role modal.

src/react/utils/index.js
```diff
@@ -40,6 +40,10 @@ export function formatDate(d) {
   return `${d.toDateString()} ${d.toTimeString().split(' ')[0]}`;
 }
 
+export function formatRoleSessionName(oidc) {
```
JBWatenbergScality: Can you please type this argument using https://github.com/scality/zenko-ui/blob/development/1.1/src/types/state.js#L39 ?

JBWatenbergScality:
> "RoleSessionName":"zenko@example.com"

It may be a bit off topic of this PR but I'm a bit concerned about logging user email in ARWWI logs. Email is a personal information that need to be coupled with specific retention policy (GDPR compliance and so on...), I would advise to better use the user id

src/react/utils/index.js
ghost: I'm not sure about this. I didn't see a requirement for the session name, but we agreed that defining it as the unique `email` identifier from the OIDC JWT claim was a good way to allow for dissociable log generation for each identity. Your comment is true though, we should not use a default value.

I propose to default to the (mandatory) `sub` claim value if the email is not present. This claim is also a unique identitier of identities within keycloak, although it is not as human-readable. I will make the needed changes so that we correctly reject the request if the `oidc.user.profile` object is not defined, otherwise we can fallback to the `sub` claim. What do you think?

Something like that:
`return ${oidc.user?.profile.email || ${oidc.user?.profile.sub};`

ghost:
Totally agree about the GDPR concerns, I will use the `sub` claim instead, it looks like this: `9afe9b78-7e9f-4944-8573-413e7d6ee285`. What do you think?

JBWatenbergScality:
> Totally agree about the GDPR concerns, I will use the `sub` claim instead, it looks like this: `9afe9b78-7e9f-4944-8573-413e7d6ee285`. What do you think?

It looks good to me !

src/react/utils/index.js
ghost: Switched to the `sub` claim. We do not need to verify that the `oidc.user.profile` exists because it is already done in the `src/react/actions/auth.js` file (please correct me if I'm wrong), so I removed the function.
See https://github.com/scality/zenko-ui/pull/203/commits/adcacbb9e7853271fc72347398fc3dc0929f24f3

Pensieve API is also updated to use the `sub` claim.

src/react/actions/sts.js
```diff
@@ -16,8 +16,8 @@ export function assumeRoleWithWebIdentity(
     const { oidc } = getState();
     const assumeRoleParams = {
       idToken: oidc.user.id_token,
-      RoleSessionName: 'app1',
       roleArn,
+      RoleSessionName: oidc.user.profile.sub,
     };
```
JBWatenbergScality: We should probably try to avoid to duplicate the "logic" here and https://github.com/scality/zenko-ui/pull/203/files#diff-8026add9d0d1ea46017507c189fdfdb7c83af900017db5106efa81c370ed518aR123 versus `getAssumeRoleWithWebIdentityIAM` in IAM client. If you can address it part of this PR that would be awesome, else we can open a debt ticket and address it later.

JBWatenbergScality (approved):
Other than the above comment that can be addressed separately the implementation looks good to me.

src/react/actions/sts.js
ghost: Done: https://github.com/scality/zenko-ui/pull/203/commits/403be56e5e96c5739c486110eda2499bc7d8360b

ghost (approved):
LGTM

src/js/IAMClient.js
```diff
@@ -1,11 +1,24 @@
 // @noflow
 
-import type { AppState } from '../types/state';
+import type { AppState, AssumeRoleParams, OIDCState } from '../types/state';
 import type { Credentials } from '../types/zenko';
 import IAM from 'aws-sdk/clients/iam';
 import type { IAMClient as IAMClientInterface } from '../types/iam';
 import { getClients } from '../react/utils/actions';
 
+export const roleName = 'scality-internal/storage-manager-role';
```
anurag4DSB: I think we should rename this to rolePath?
Or specify, the internal path it `scality-internal` and `storage-manager-role` is the name.

anurag4DSB (commented):
LGTM, just a minor comment.

src/js/IAMClient.js
```diff
@@ -1,11 +1,24 @@
...
+
+export function getAssumeRoleWithWebIdentityParams(
+  oidc: OIDCState,
+  accountID: string,
+): AssumeRoleParams {
+  return {
+    idToken: oidc.user.id_token,
+    roleArn: `arn:aws:iam::${accountID}:role/${roleName}`,
+    RoleSessionName: oidc.user.profile.sub,
```
anurag4DSB, line 18: I understand the GDPR concerns, but it was decided to use email for auditing purposes and identifying the user.
Would you be able to create a documentation ticket on how we can find identity of a user from the Sub(session name)?
You can link it as a blocker for https://scality.atlassian.net/browse/ARTESCA-2553

ghost: The sub claim is the user unique identifier `<user_id>` from keycloak. To get the user information from this ID, we can either go to 
`<keycloak_url>/auth/admin/master/console/#/realms/<realm_name>/users/<user_id>` or, if we want to get it programmatically, we can use keycloak's API: [https://www.keycloak.org/docs-api/11.0/rest-api/index.html#_getuser](https://www.keycloak.org/docs-api/11.0/rest-api/index.html#_getuser).

The identity is indeed an email as per the design, but there is no clear mapping between the use of this email in the logs and how to identify the session initiator: `To assist in tracing the 'origin' of such requests, the RoleSessionName is provided to identify the particular assumption.`. The user id is opaque and can (should) be used to identify the user while complying with GDPR.

Example:
```
eyJhbGciOiJSUzI1NiIsInR5cCIgOiAiSldUIiwia2lkIiA6ICJZNUU4LW52b1RZeE43ekpEVktOT1lKWnFzd2dVSWZJOFJLOXJTTExZQTNvIn0.eyJleHAiOjE2MzgxNzE3NjIsImlhdCI6MTYzODE3MTQ2MiwianRpIjoiYmMyZGQzZGItYzE1YS00N2YxLThhZDYtZDVkNGM3NmIyNTc5IiwiaXNzIjoiaHR0cHM6Ly8xMjcuMC4wLjE6ODQ0My9hdXRoL3JlYWxtcy9teXJlYWxtIiwiYXVkIjoiYWNjb3VudCIsInN1YiI6ImE1YTc0NTUyLTJjMWYtNDk5ZC1iZmNkLWYzN2FlMDBiMTM1YyIsInR5cCI6IkJlYXJlciIsImF6cCI6Im15Y2xpZW50Iiwic2Vzc2lvbl9zdGF0ZSI6IjI2MWQ2MjhiLTcxOGMtNDZiZC05M2NkLTgzMWU2NTQ5ZjczNSIsImFjciI6IjEiLCJyZWFsbV9hY2Nlc3MiOnsicm9sZXMiOlsib2ZmbGluZV9hY2Nlc3MiLCJTdG9yYWdlTWFuYWdlciIsInVtYV9hdXRob3JpemF0aW9uIl19LCJyZXNvdXJjZV9hY2Nlc3MiOnsiYWNjb3VudCI6eyJyb2xlcyI6WyJtYW5hZ2UtYWNjb3VudCIsIm1hbmFnZS1hY2NvdW50LWxpbmtzIiwidmlldy1wcm9maWxlIl19fSwic2NvcGUiOiJwcm9maWxlIGVtYWlsIiwiZW1haWxfdmVyaWZpZWQiOnRydWUsInJvbGVzIjpbIm9mZmxpbmVfYWNjZXNzIiwiU3RvcmFnZU1hbmFnZXIiLCJ1bWFfYXV0aG9yaXphdGlvbiJdLCJwcmVmZXJyZWRfdXNlcm5hbWUiOiJjYWU5MWU0NWFlZDgwZjNhM2ZlMjg1YzNjOGMxYTdlNzhkODJkNDczIiwiZW1haWwiOiJjYWU5MWU0NWFlZDgwZjNhM2ZlMjg1YzNjOGMxYTdlNzhkODJkNDczQHRlc3QuY29tIiwiZ3JvdXAiOltdfQ.utE8nC1nKg6g3NM5GxnmuAGulWtiSwHYNcMoirzxx_UAkYUbrkCI-azKJfsEjF_fkp4g1ACWsNiMUjtiVmq_kO-qLqAcvk0DYZmvNMEUV6u28SVHvSA_2N-6xFI4Oxo9hhEZ9FSc-WWji3qJVXuB3P3o1gR36I-qnIhLo6RLap8e-S90khBAMX2lnNuDpsfGeEYET_dMWJ2ZvzikwdJK5UNhAYl4RUo-kRcataoCxcSHbAmMebaVCn52VHMl5j-PCV2gZt6KXL_gF-k4vI1vsX0_lVu2z8q-WyyzdD-J_w6C7qXH8Hqb9W-VVHJMBUmAHobFwspjrAVh3nZjia1_mA
```

Wille have the following sub claim: `"sub": "a5a74552-2c1f-499d-bfcd-f37ae00b135c",`

Later on, we can find the user in keycloak using this id:
image

src/js/IAMClient.js
ghost: I will rename it `rolePathName` as we have both.

src/js/IAMClient.js
ghost: Ticket added.

src/js/IAMClient.js
ghost: Done: https://github.com/scality/zenko-ui/commit/0aa0982448282802472adc30314a3164335a4b0f#diff-524d780d05f66cc8206f06f1bb1dd913e4932608ba6d60a6be5cf1f746113ba7R9

src/js/IAMClient.js
ghost: By the way, it is still possible for the administrator, if he/she wants to, to set the `sub` claim value to be the user's email.

anurag4DSB: cool, I agree with your approach, just making sure we document that.

anurag4DSB (approved):

src/js/IAMClient.js
nicolas2bert: @williamlardier It looks like the role session name includes the service name in Pensieve Api. Should we follow the same logic for zenko-ui project? https://github.com/scality/pensieve-api/pull/143/files#diff-fa3ccffbf25e42d6d0eeba977259607da1812212f3df390d2c6a0523e8018bffR63

ghost: For me the idea was to prefix the service name when it wasn't ZKUI, but I don't see any problem with adding zenko-ui here as well. If you think it's a good thing to do, let me know, I'll reopen a PR with this change!

PR #204, opened 2021-11-23
ZKUI-35 location creation - phrasing

nicolas2bert:
Video of the Async Notification feature:

https://scality.atlassian.net/jira/software/c/projects/OS/boards/215?modal=detail&selectedIssue=OS-57

### DIFF
--- a/src/react/actions/__tests__/zenko.test.js
+++ b/src/react/actions/__tests__/zenko.test.js
@@ -29,7 +29,7 @@ const continueSearchNetworkAction = dispatchAction.NETWORK_START_ACTION(
   NETWORK_START_ACTION_CONTINUE_SEARCH,
 );
 const pauseIngestionSiteNetworkAction = dispatchAction.NETWORK_START_ACTION(
-  'Pausing ingestion',
+  'Pausing Async Notification',
 );
 
 describe('zenko actions', () => {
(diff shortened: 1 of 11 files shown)


nicolas2bert:
ping

miniscruff (approved):

Cuervino (changes requested):
image
1. Checkbox is not perfectly aligned.

2. All the tooltips should have their content aligned to the left.

3. Why do you need to create a Bucket with Async Notification Ready and Versioning (2:05)? Is Versioning mandatory? 
In this case, maybe adding a automatic toggling ON logic when Async Notification Ready is selected.

4. The status of the Async Notification shift from "-" to "Paused (?)" after few seconds (2:21). 
If the status has to be retrieved, then it should be first a "loading" status, to explain that the information is not yet available.
Currently, there is no way for the users to understand that the status they see at first will be shortly updated.

nicolas2bert:
1. core-ui: ok on smaller screen - checkbox should grow linearly with the size of font.
2. core-ui: https://github.com/scality/core-ui/blob/development/1.0/src/lib/components/tooltip/Tooltip.component.js#L46 ?
3. No need to be versioned. Recommended.
4. https://scality.atlassian.net/browse/ZKUI-41

Cuervino (approved):

nicolas2bert:
@bert-e approve

PR #208, opened 2021-12-07
ZKUI-29 - Fix reauth modal popup behavior

MonPote:
The issue:
When the user switch tab or leave the laptop for while, the reauth
modal will popup but the behavior is not clear and it confuse the
user.

We identify 2 issues:

1. A race condition
When the user go back to the tab, the UI will trigger both reauth
(because the token might be expired) and refresh the instance
(See the `useEffect` in `Routes.jsx`).
This will result a 401 and display the popup even though the token
is refresh right after.

2. A token expired for too long which require a relogin Keycloak.
In this case if the user try to click on `Reload` is will try to
reauth but, it will fail and the user will be stuck until the page
is refreshed.

To address both issues, we choose to :
- check the expired time in the polling.
This prevent to display a 401 right after the user come back from
a new tab.
- add `ADD_OIDC_USER` and `LOAD_CLIENTS_SUCCESS` to remove modal
- only log out or reload the entire page if the modal appear

Finally, this fix is a workaround before we migrate to Module
Federation and react query.

### DIFF
--- a/src/react/ZenkoUI.js
+++ b/src/react/ZenkoUI.js
@@ -21,6 +21,7 @@ function ZenkoUI() {
   const isConfigLoaded = useSelector(
     (state: AppState) => state.auth.isConfigLoaded,
   );
+
   const configFailure = useSelector(
     (state: AppState) => state.auth.configFailure,
   );
--- a/src/react/ui-elements/ReauthDialog.jsx
+++ b/src/react/ui-elements/ReauthDialog.jsx
@@ -1,11 +1,9 @@
 // @flow
-import { loadClients, networkAuthReset } from '../actions';
-import { useDispatch, useSelector } from 'react-redux';
+import { useSelector } from 'react-redux';
 import type { AppState } from '../../types/state';
 import { Button } from '@scality/core-ui/dist/next';
 import { CustomModal as Modal } from './Modal';
 import React from 'react';
-import { push } from 'connected-react-router';
 import { spacing } from '@scality/core-ui/dist/style/theme';
 
 const DEFAULT_MESSAGE = 'We need to log you in.';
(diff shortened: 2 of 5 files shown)


src/react/ui-elements/ReauthDialog.jsx
```diff
@@ -17,26 +15,16 @@ const ReauthDialog = () => {
...
-
   if (!needReauth) {
     return null;
   }
 
   return (
     <Modal
       id="reauth-dialog-modal"
-      close={() => reauth(pathname)}
+      close={() => window.location.reload()}
```
gaudiauj, line 27:
```suggestion
      close={window.location.reload}
```

src/react/ui-elements/ReauthDialog.jsx
```diff
@@ -50,7 +38,7 @@ const ReauthDialog = () => {
           )}
           <Button
             variant="primary"
-            onClick={() => reauth(pathname)}
+            onClick={() => window.location.reload()}
```
gaudiauj, line 41:
```suggestion
            onClick={window.location.reload}
```

src/react/ZenkoUI.js
```diff
@@ -21,19 +21,27 @@ function ZenkoUI() {
...
+  const [configLoaded, setConfigLoaded] = useState(isConfigLoaded);
 
   const dispatch = useDispatch();
 
   useEffect(() => {
     dispatch(loadAppConfig());
-  }, [dispatch]);
+  }, []);
+
+  useEffect(() => {
```
gaudiauj: for me the `isConfigLoaded` is sufficient, since it come from a reactive context (redux) it should be updated, and force the render everytime it change without the need to use useEffect and useState.

MonPote: Good catch

src/react/ui-elements/ReauthDialog.jsx
MonPote: I will keep it this way because it will cause a :
```
Uncaught TypeError: Illegal invocation react
```

ChengYanJin (approved):
LGTM!

gaudiauj (approved):
approved

PR #210, opened 2021-12-14
ZKUI-29 - Fix reauth modal popup behavior

MonPote:
The issue:
When the user switch tab or leave the laptop for while, the reauth
modal will popup but the behavior is not clear and it confuse the
user.

We identify 2 issues:

1. A race condition
When the user go back to the tab, the UI will trigger both reauth
(because the token might be expired) and refresh the instance
(See the `useEffect` in `Routes.jsx`).
This will result a 401 and display the popup even though the token
is refresh right after.

2. A token expired for too long which require a relogin Keycloak.
In this case if the user try to click on `Reload` is will try to
reauth but, it will fail and the user will be stuck until the page
is refreshed.

To address both issues, we choose to :
- check the expired time in the polling.
This prevent to display a 401 right after the user come back from
a new tab.
- add `ADD_OIDC_USER` and `LOAD_CLIENTS_SUCCESS` to remove modal
- only log out or reload the entire page if the modal appear

Finally, this fix is a workaround before we migrate to Module
Federation and react query.

### DIFF
--- a/src/react/Routes.jsx
+++ b/src/react/Routes.jsx
@@ -35,14 +35,18 @@ function PrivateRoutes() {
       // That will fix management API request being canceled during autentication.
       dispatch(loadClients());
 
-      const refreshIntervalStatsUnit = setInterval(
-        () => dispatch(loadInstanceLatestStatus()),
-        10000,
-      );
-      const refreshIntervalStatsSeries = setInterval(
-        () => dispatch(loadInstanceStats()),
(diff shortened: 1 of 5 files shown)


williamlardier:
ping

MonPote:
ping

JBWatenbergScality (approved):

ghost (approved):

PR #214, opened 2021-12-28
ZKUI-53 & ZKUI-54: Add legal hold in the object properties

ChengYanJin:
ZKUI-53: Add balance-scale in the object list
ZKUI-54: Add legal hold toggle in the object properties

image
image

### DIFF
--- a/src/react/actions/s3object.js
+++ b/src/react/actions/s3object.js
@@ -114,6 +114,7 @@ export function getObjectMetadataSuccess(
     Mode: RetentionMode,
     RetainUntilDate: Date,
   |},
+  isLegalHoldEnabled: boolean,
 ): GetObjectMetadataSuccessAction {
   return {
     type: 'GET_OBJECT_METADATA_SUCCESS',
@@ -122,6 +123,7 @@ export function getObjectMetadataSuccess(
     info,
     tags,
     ObjectRetention,
+    isLegalHoldEnabled,
   };
 }
 
@@ -231,6 +233,11 @@ function _getObjectListWithSignedUrlAndLockStatus<
           object.VersionId,
         ),
         ObjectRetention: await zenkoClient.getObjectRetention(
+          bucketName,
+          object.Key,
+          object.VersionId,
+        ),
+        IsLegalHoldEnabled: await zenkoClient.getObjectLegalHold(
           bucketName,
           object.Key,
           object.VersionId,
@@ -480,8 +487,9 @@ export function getObjectMetadata(
       zenkoClient.headObject(bucketName, objectKey, versionId),
       zenkoClient.getObjectTagging(bucketName, objectKey, versionId),
       zenkoClient.getObjectRetention(bucketName, objectKey, versionId),
+      zenkoClient.getObjectLegalHold(bucketName, objectKey, versionId),
     ])
-      .then(([info, tags, objectRetention]) =>
+      .then(([info, tags, objectRetention, isLegalHoldEnabled]) =>
         dispatch(
           getObjectMetadataSuccess(
             bucketName,
@@ -489,6 +497,7 @@ export function getObjectMetadata(
             info,
             tags.TagSet,
             objectRetention,
+            isLegalHoldEnabled,
           ),
         ),
       )
@@ -527,6 +536,26 @@ export function putObjectTagging(
     dispatch(networkStart('Getting object tags'));
     return zenkoClient
       .putObjectTagging(bucketName, objectKey, tags, versionId)
+      .then(() => dispatch(getObjectMetadata(bucketName, objectKey, versionId)))
+      .catch(error => dispatch(handleAWSClientError(error)))
+      .catch(error => dispatch(handleAWSError(error, 'byModal')))
+      .finally(() => dispatch(networkEnd()));
+  };
+}
+
+export function putObjectLegalHold(
+  bucketName: string,
+  objectKey: string,
+  versionId: string,
+  isLegalHold: boolean,
+  prefixWithSlash: string,
+): ThunkStatePromisedAction {
+  return (dispatch, getState) => {
+    const { zenkoClient } = getClients(getState());
+    dispatch(networkStart('Getting object legal hold'));
+    return zenkoClient
+      .putObjectLegalHold(bucketName, objectKey, versionId, isLegalHold)
+      .then(() => dispatch(listObjects(bucketName, prefixWithSlash)))
       .then(() => dispatch(getObjectMetadata(bucketName, objectKey, versionId)))
       .catch(error => dispatch(handleAWSClientError(error)))
       .catch(error => dispatch(handleAWSError(error, 'byModal')))
--- a/src/react/databrowser/objects/details/Properties.jsx
+++ b/src/react/databrowser/objects/details/Properties.jsx
@@ -1,74 +1,141 @@
 // @noflow
-import Table, * as T from '../../../ui-elements/TableKeyValue';
+import Table, * as T from '../../../ui-elements/TableKeyValue2';
 import { Clipboard } from '../../../ui-elements/Clipboard';
 import MiddleEllipsis from '../../../ui-elements/MiddleEllipsis';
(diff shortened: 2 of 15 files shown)


src/react/actions/s3object.js
```diff
@@ -533,3 +543,23 @@ export function putObjectTagging(
...
+  isLegalHold: boolean,
+  prefixWithSlash: string,
+): ThunkStatePromisedAction {
+  return (dispatch, getState) => {
+    const { zenkoClient } = getClients(getState());
+    dispatch(networkStart('Getting object legal hold'));
+    return zenkoClient
+      .putObjectLegalHold(bucketName, objectKey, versionId, isLegalHold)
+      .then(() => dispatch(getObjectMetadata(bucketName, objectKey, versionId)))
+      .then(() => dispatch(listObjects(bucketName, prefixWithSlash)))
```
JBWatenbergScality: Maybe we could just dispatch the action that select the object here temporarily while the url rework is not done ?

src/react/databrowser/objects/details/Properties.jsx
```diff
@@ -17,6 +19,10 @@ const TruncatedValue = styled(T.Value)`
 `;
 
 function Properties({ objectMetadata }: Props) {
+  const dispatch = useDispatch();
+  const prefixWithSlash = usePrefixWithSlash();
+  const isLegalHoldEnabled = objectMetadata.legalHoldStatus === 'ON';
```
JBWatenbergScality: I'm wondering wether we should not keep this ON logic on the client side and everywhere else use a boolean.
In src/js/S3Client.js we could change :
```
getObjectLegalHold(bucketName, objectKey, versionId) {
    const params = {
      Bucket: bucketName,
      Key: objectKey,
      VersionId: versionId,
    };

    return new Promise((resolve, reject) => {
      this.client.getObjectLegalHold(params, (error, data) => {
        if (error) {
          if (
            error.code === 'NoSuchObjectLockConfiguration' ||
            error.code === 'InvalidRequest'
          ) {
            return resolve('OFF');
          }
          return reject(error);
        }
        if (data.Status) {
          return resolve(data.Status);
        }
        return resolve('OFF');
      });
    });
  }
```

To 
```
getObjectLegalHold(bucketName, objectKey, versionId) {
    const params = {
      Bucket: bucketName,
      Key: objectKey,
      VersionId: versionId,
    };

    return new Promise((resolve, reject) => {
      this.client.getObjectLegalHold(params, (error, data) => {
        if (error) {
          if (
            error.code === 'NoSuchObjectLockConfiguration' ||
            error.code === 'InvalidRequest'
          ) {
            return resolve(false);
          }
          return reject(error);
        }
        if (data.Status) {
          return resolve(data.Status === 'ON');
        }
        return resolve(false);
      });
    });
  }
```

Then everywhere else we wouldn't have to take care of the boolean translation of the value

src/react/databrowser/objects/details/Properties.jsx
```diff
@@ -61,14 +67,53 @@ function Properties({ objectMetadata }: Props) {
...
             <T.Key> Lock </T.Key>
             <T.Value>
-              {objectMetadata.lockStatus === 'LOCKED' && <>Locked ({objectMetadata.objectRetention.mode.toLowerCase()})<br />until {objectMetadata.objectRetention.retainUntilDate}</>}
-              {objectMetadata.lockStatus === 'RELEASED' && `Released - since ${objectMetadata.objectRetention.retainUntilDate}`}
+              {objectMetadata.lockStatus === 'LOCKED' && (
+                <>
+                  Locked ({objectMetadata.objectRetention.mode.toLowerCase()})
+                  <br />
+                  until {objectMetadata.objectRetention.retainUntilDate}
+                </>
```
JBWatenbergScality: I think it is missing the locked icon here for the discoverability of the icon meaning (which was the starting point of this refactoring

src/react/databrowser/objects/details/Properties.jsx
```diff
@@ -61,14 +67,53 @@ function Properties({ objectMetadata }: Props) {
...
+              )}
+              {objectMetadata.lockStatus === 'RELEASED' &&
+                `Released - since ${objectMetadata.objectRetention.retainUntilDate}`}
```
JBWatenbergScality: I think the unlocked icon is also missing here

src/react/databrowser/objects/details/Properties.jsx
```diff
@@ -61,14 +67,53 @@ function Properties({ objectMetadata }: Props) {
...
+                        !isLegalHoldEnabled,
+                        prefixWithSlash,
+                      ),
+                    )
+                  }
+                />
+                {isLegalHoldEnabled && (
+                  <i
+                    className="fas fa-balance-scale"
+                    style={{ paddingLeft: '5px' }}
```
JBWatenbergScality: shouldn't we use 0.5rem instead ? Maybe zenko is not yet ready for the rem unit conversion, I'm just wonering

src/react/databrowser/objects/details/Properties.jsx
```diff
@@ -1,5 +1,5 @@
 // @noflow
-import Table, * as T from '../../../ui-elements/TableKeyValue';
+import Table, * as T from '../../../ui-elements/TableKeyValue2';
```
JBWatenbergScality, line 2: Is TableKeyValue still used after this change ? If not I think we could drop it and rename the TableKeyValue2 to TableKeyValue 🚀

src/react/databrowser/objects/details/Properties.jsx
JBWatenbergScality: I've seen it afterwards in commit https://github.com/scality/zenko-ui/pull/214/commits/540df519b27ffa9a9e66cf2d34a0a85fdb9afd9b

src/react/databrowser/objects/details/Properties.jsx
JBWatenbergScality: I've seen it afterwards in commit https://github.com/scality/zenko-ui/pull/214/commits/540df519b27ffa9a9e66cf2d34a0a85fdb9afd9b

src/react/databrowser/objects/details/Properties.jsx
ChengYanJin: Sadly this TableKeyValue is still used by many other components :(

src/react/actions/s3object.js
ChengYanJin: that's right!!! I will just reverse L.559 and L.560.  🤩  
and problem solved!

src/react/databrowser/objects/details/Properties.jsx
ChengYanJin: I actually changed in the later commit with `spacing`
```
const Icon = styled.i`
  margin-left: ${spacing.sp4};
`;
```

src/react/databrowser/objects/details/Properties.jsx
ChengYanJin: Yeah. It makes sense to store the boolean `isLegalHoldEnabled`

src/react/databrowser/objects/details/Properties.jsx
JBWatenbergScality: Would be for next time then !

JBWatenbergScality (approved):

src/react/databrowser/objects/details/Properties.jsx
```diff
@@ -1,74 +1,141 @@
...
   objectMetadata: ObjectMetadata,
 };
 
-const TruncatedValue = styled(T.Value)`
-  max-width: 300px;
+const TruncatedValue = styled.div`
+  max-width: 18rem;
+`;
+
+const Icon = styled.i`
```
nicolas2bert, line 23: NIT: I know ui-elements/Table as an `Icon` component you might want to reuse. Or you can introduce one in `TableKeyValue2`

nicolas2bert (approved):

PR #215, opened 2022-01-03
ZKUI 57-Add query params in the url

ChengYanJin:
Add the following props in the query params:
- showversions
- prefix
- versionId
- metadatasearch

With this PR, we centralize the  **listObject** action only in one place.

Note that this PR will change the behaviors:
- We will keep the current object selection when toggling the list versions.

Note that:
**We bumb the core-ui version with v0.28.0**

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -12,3 +12,11 @@ export function chunkArray(
 
   return tempArray;
 }
+
+export function removeTrailingSlash(prefix: string): string {
+  if (prefix.slice(-1) === '/') {
+    return prefix.slice(0, -1);
+  } else {
+    return prefix;
+  }
+}
--- a/src/react/databrowser/objects/MetadataSearch.jsx
+++ b/src/react/databrowser/objects/MetadataSearch.jsx
@@ -48,8 +48,8 @@ type Props = {
   errorZenkoMsg: ?string,
 };
 const MetadataSearch = ({ isMetadataType, errorZenkoMsg }: Props) => {
-  const [inputText, setInputText] = useState('');
   const [hintsShown, setHintsShown] = useState(false);
+
   const dispatch: DispatchAPI<Action> = useDispatch();
   const query = useQuery();
   const { pathname } = useLocation();
@@ -58,6 +58,8 @@ const MetadataSearch = ({ isMetadataType, errorZenkoMsg }: Props) => {
   const suggestionsRef = useRef(null);
   const inputRef = useRef<SearchMetadataInput<> | null>(null);
   useOutsideClick(suggestionsRef, () => setHintsShown(false));
+  const searchInput = query.get('metadatasearch');
+  const [inputText, setInputText] = useState(searchInput || '');
 
   const handleChange = e => {
     setInputText(e.target.value);
@@ -99,7 +101,7 @@ const MetadataSearch = ({ isMetadataType, errorZenkoMsg }: Props) => {
 
   return (
     <SearchMetadataContainer
-      isHidden={prefixWithSlash ? 1 : 0}
+      isHidden={prefixWithSlash && !searchInput ? 1 : 0}
       onSubmit={handleSubmit}
     >
       <SearchMetadataInputAndIcon>
--- a/src/react/databrowser/objects/MetadataSearch.jsx
+++ b/src/react/databrowser/objects/MetadataSearch.jsx
@@ -9,11 +9,16 @@ import {
   SearchMetadataInputAndIcon,
   SearchValidationIcon,
 } from '../../ui-elements/Table';
-import { listObjects, newSearchListing } from '../../actions';
 import type { Action } from '../../../types/actions';
 import type { DispatchAPI } from 'redux';
 import { useDispatch } from 'react-redux';
-import { useOutsideClick } from '../../utils/hooks';
+import {
+  useOutsideClick,
+  useQuery,
+  usePrefixWithSlash,
+} from '../../utils/hooks';
+import { useLocation } from 'react-router';
+import { push } from 'connected-react-router';
 
 export const METADATA_SEARCH_HINT_ITEMS = [
   { descr: 'files with extension ".pdf"', q: 'key like /pdf$/' },
@@ -39,21 +44,16 @@ export const METADATA_SEARCH_HINT_ITEMS = [
 ];
 
 type Props = {
-  bucketName: string,
-  prefixWithSlash: string,
   isMetadataType: boolean,
   errorZenkoMsg: ?string,
 };
-const MetadataSearch = ({
-  isMetadataType,
-  bucketName,
-  prefixWithSlash,
-  errorZenkoMsg,
-}: Props) => {
+const MetadataSearch = ({ isMetadataType, errorZenkoMsg }: Props) => {
   const [inputText, setInputText] = useState('');
   const [hintsShown, setHintsShown] = useState(false);
   const dispatch: DispatchAPI<Action> = useDispatch();
-
+  const query = useQuery();
+  const { pathname } = useLocation();
+  const prefixWithSlash = usePrefixWithSlash();
   // hide hints if clicked on outside of element.
   const suggestionsRef = useRef(null);
   const inputRef = useRef<SearchMetadataInput<> | null>(null);
@@ -68,14 +68,18 @@ const MetadataSearch = ({
     if (!inputText || prefixWithSlash) {
       return;
     }
-    dispatch(newSearchListing(bucketName, inputText));
+    // Add metadatasearch in the query params
+    query.set('metadatasearch', inputText);
+    dispatch(push(`${pathname}?${query.toString()}`));
   };
 
   const reset = e => {
     e.stopPropagation();
     setHintsShown(false);
     setInputText('');
-    dispatch(listObjects(bucketName, prefixWithSlash));
+    // Remove the medatasearch from the query params
+    query.delete('metadatasearch');
+    dispatch(push(`${pathname}?${query.toString()}`));
   };
 
   const handleHintClicked = q => {
--- a/src/react/databrowser/objects/ObjectDetails.jsx
+++ b/src/react/databrowser/objects/ObjectDetails.jsx
@@ -1,13 +1,13 @@
 // @flow
+import React from 'react';
+import type { Node } from 'react';
 import type { AppState } from '../../../types/state';
 import { ContentSection } from '../../ui-elements/ListLayout2';
 import { CustomTabs } from '../../ui-elements/Tabs';
 import { List } from 'immutable';
 import type { ListObjectsType } from '../../../types/s3';
 import Metadata from './details/Metadata';
-import type { Node } from 'react';
 import Properties from './details/Properties';
-import React from 'react';
 import Tags from './details/Tags';
 import { Warning } from '../../ui-elements/Warning';
 import { useLocation } from 'react-router-dom';
@@ -46,8 +46,8 @@ function ObjectDetails({ toggled, listType }: Props) {
   const objectMetadata = useSelector(
     (state: AppState) => state.s3.objectMetadata,
   );
-
   const tabName = query.get('tab');
+  const queryObject = Object.fromEntries(query.entries());
 
   const details = () => {
     if (toggled.size > 1) {
@@ -71,17 +71,25 @@ function ObjectDetails({ toggled, listType }: Props) {
   return (
     <ContentSection>
       <CustomTabs>
-        <CustomTabs.Tab label="Summary" path={pathname}>
+        <CustomTabs.Tab
+          label="Summary"
+          path={pathname}
+          query={{ ...queryObject, ...{ tab: null } }}
+        >
           {details()}
         </CustomTabs.Tab>
         <CustomTabs.Tab
           label="Metadata"
           path={pathname}
-          query={{ tab: 'metadata' }}
+          query={{ ...queryObject, ...{ tab: 'metadata' } }}
         >
           {details()}
         </CustomTabs.Tab>
-        <CustomTabs.Tab label="Tags" path={pathname} query={{ tab: 'tags' }}>
+        <CustomTabs.Tab
+          label="Tags"
+          path={pathname}
+          query={{ ...queryObject, ...{ tab: 'tags' } }}
+        >
           {details()}
         </CustomTabs.Tab>
       </CustomTabs>
--- a/src/react/databrowser/objects/ObjectList.jsx
+++ b/src/react/databrowser/objects/ObjectList.jsx
@@ -9,12 +9,10 @@ import type {
 } from '../../../types/s3';
 import {
   LIST_OBJECTS_METADATA_TYPE,
-  LIST_OBJECTS_S3_TYPE,
   LIST_OBJECT_VERSIONS_S3_TYPE,
 } from '../../utils/s3';
 import { isVersioningDisabled, maybePluralize } from '../../utils';
 import {
-  listObjects,
   openFolderCreateModal,
   openObjectDeleteModal,
   openObjectUploadModal,
@@ -27,6 +25,9 @@ import ObjectListTable from './ObjectListTable';
 import React from 'react';
 import { Toggle } from '@scality/core-ui';
 import { WarningMetadata } from '../../ui-elements/Warning';
+import { push } from 'connected-react-router';
+import { useQuery } from '../../utils/hooks';
+import { useLocation } from 'react-router';
 
 type Props = {
   objects: List<ObjectEntity>,
@@ -46,6 +47,9 @@ export default function ObjectList({
   bucketInfo,
 }: Props) {
   const dispatch = useDispatch();
+  const { pathname } = useLocation();
+  const query = useQuery();
+
   const errorZenkoMsg = useSelector(
     (state: AppState) => state.zenko.error.message,
   );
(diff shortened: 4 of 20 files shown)


src/react/databrowser/objects/ObjectList.jsx
```diff
@@ -114,6 +120,8 @@ export default function ObjectList({
             toggle={isVersioningType}
             label="List Versions"
             onChange={() => {
+              query.set('showversions', !isVersioningType ? 'true' : 'false');
+              dispatch(push(`${pathname}?${query.toString()}`));
               const type = isVersioningType
                 ? LIST_OBJECTS_S3_TYPE
                 : LIST_OBJECT_VERSIONS_S3_TYPE;
```
JBWatenbergScality: here and the line just below, do we still need to dispatch listObjects ? Can't we just rely on the fact that the url changing the Objects component will trigger the listing of the objects ?

src/react/databrowser/objects/ObjectRow.jsx
```diff
@@ -58,9 +61,17 @@ const Row = ({
   const row = rows[index];
   prepareRow(row);
 
+  const { pathname } = useLocation();
+  const query = useQuery();
+
   const handleClick = () => {
     dispatch(toggleAllObjects(false));
     dispatch(toggleObject(row.original.key, row.original.versionId));
```
JBWatenbergScality: I think this line may be dropped as the ObjectListTable will dispatch it when the url change

src/react/databrowser/objects/ObjectListTable.jsx
```diff
@@ -78,6 +79,14 @@ export default function ObjectListTable({
     [dispatch],
   );
 
+  const query = useQuery();
+  const objectKey = query.get('prefix');
+  const versionId = query.get('versionId');
+
+  useEffect(() => {
+    dispatch(toggleObject(objectKey, versionId ? versionId : undefined));
```
JBWatenbergScality: I see in the ObjectRow that before togglingAnObject it is dispatching `dispatch(toggleAllObjects(false));` do we nned to do so here as well ? What does `dispatch(toggleAllObjects(false));` exactly do ? I guess it is dis-selecting all the line of the table, if so I think we should likely not do that as it would loose the selection context of the user as soon as the user is checking one of the object _they_ selected

src/react/databrowser/objects/MetadataSearch.jsx
```diff
@@ -69,13 +73,19 @@ const MetadataSearch = ({
       return;
     }
     dispatch(newSearchListing(bucketName, inputText));
```
JBWatenbergScality: I think this line may be dropped ?

src/react/databrowser/objects/MetadataSearch.jsx
```diff
@@ -69,13 +73,19 @@ const MetadataSearch = ({
...
+    // Add metadatasearch in the query params
+    query.set('metadatasearch', inputText);
+    dispatch(push(`${pathname}?${query.toString()}`));
   };
 
   const reset = e => {
     e.stopPropagation();
     setHintsShown(false);
     setInputText('');
     dispatch(listObjects(bucketName, prefixWithSlash));
```
JBWatenbergScality: I think this line may be dropped ?

src/react/databrowser/objects/ObjectListTable.jsx
```diff
@@ -83,11 +83,22 @@ export default function ObjectListTable({
...
   useEffect(() => {
-    dispatch(toggleObject(objectKey, versionId ? versionId : undefined));
-  }, [dispatch]);
+    dispatch(
+      toggleObject(
+        objectKey,
+        versionId && isListVersion ? versionId : undefined,
+      ),
+    );
+    // Add `objectsLength` as dependency to make sure the toogleObject action is triggered after listObject
```
JBWatenbergScality: Then I guess we should wrap the above line with an `if (objectsLength) {}` so that it is more explicit and we don't need this comment to explain why we need that dependency

src/react/databrowser/objects/ObjectList.jsx
JBWatenbergScality: addressed by https://github.com/scality/zenko-ui/pull/215/commits/cf36e256ad62db4dfff8273954f3a2caadee60fb I guess

src/react/databrowser/objects/ObjectListTable.jsx
ChengYanJin: Good point !!

src/react/databrowser/objects/ObjectListTable.jsx
ChengYanJin: The issue with removing the `dispatch(toggleAllObjects(false));` is that given an object is already selected, it would not be possible now to check another object, because as soon as I click on another object, its a multiple selection state :(

src/react/databrowser/objects/MetadataSearch.jsx
ChengYanJin: 🤔  I don't think we should drop this line, it is the action to do the search objects.

src/react/databrowser/objects/MetadataSearch.jsx
ChengYanJin: 🤔  I think if we really want to drop this line, then we probably need to add the value of metadatasearch query param in the dependency list where we dispatch the `listObjects` in the Objects.jsx

I am wondering, it is something we want?

```js
  useEffect(() => {
    dispatch(
      listObjects(
        bucketNameParam,
        prefixWithSlash,
        isShowVersions ? LIST_OBJECT_VERSIONS_S3_TYPE : LIST_OBJECTS_S3_TYPE,
      ),
    ).finally(() => setLoaded(true));
  }, [bucketNameParam, prefixWithSlash, dispatch, isShowVersions]);
  ```

Comment (ChengYanJin) [reply src/react/databrowser/objects/MetadataSearch.jsx]: OK. So we took the decision to centralize the listObjects only in one place (Objects.jsx)

File: src/react/databrowser/objects/ObjectListTable.jsx [hunk 7]
```diff
@@ -94,10 +94,13 @@ export default function ObjectListTable({
...
-    // We exclude `versionId` from the dependency array,
-    // because versionId will be patched to the query params in the property tab.
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [dispatch, isListVersion, objectsLength, objectKey]);
+  }, [
+    dispatch,
+    isListVersion,
+    objectsLength,
+    objectKey,
+    isListVersion ? versionId : null,
```
JBWatenbergScality: `isListVersion, versionId`

ChengYanJin: OK. I think it would be easier to understand in this way.
Actually, we need to add a conditional dependency `versionId` in the dependency array.

The reason is that we are not listing versions, `versionId` should not appear in the dependency array. Otherwise when we patch the `versionId` to the URL query params it will trigger another **toggleObject** action with the versionId, where we don't have this object in the list..

```js
  const versionIdOrUndefined =
    isListVersion && versionId ? versionId : undefined;
  useEffect(() => {
    if (objectsLength > 0) {
      dispatch(
        toggleObject(
          isFolder ? `${objectKey}/` : objectKey,
          versionIdOrUndefined,
        ),
      );
    }
  }, [
    dispatch,
    isListVersion,
    objectsLength,
    objectKey,
    isFolder,
    versionIdOrUndefined,
  ]);
 ```

Comment (JBWatenbergScality) [reply src/react/databrowser/objects/ObjectListTable.jsx]: Ok I see, I like this approach !

**Review by JBWatenbergScality — APPROVED** [review]

**Comment by nicolas2bert** [comment]
Have not look at the code yet but I run the branch locally and found an issue with adding an object in a prefix (new folder). 
Also, it looks like we lose the context from the breadcrumb when moving into a prefix/folder.
When versioning is on, it looks like adding a new folder throws: `(Uncaught TypeError: Cannot read properties of undefined (reading 'length')`.
Also when select an object and click to Metadata or Tag tab on the right side, it looks like the query params get overwritten.

**Comment by nicolas2bert** [comment]
Some of the issue I listed might not be related to your changes and we will need to create ticket for them if not.
(This one might be out of the scope) When I reload https://<UI>/buckets/bucket-test/objects?metadatasearch=key+like+%2Fjpg%24%2F), I was expecting to see `key like /jpg$/` value set in the metadata search input.
One idea could be that, before reviewing this code, we ask @Cuervino to review the UX to see if it aligns with the requirements. 
What do you think @ChengYanJin ?

File: src/js/utils.js [hunk 8]
```diff
@@ -20,3 +21,11 @@ export function removeTrailingSlash(prefix: string): string {
...
   }
 }
+
+export function paramsToObject(query: Object) {
+  const result = {};
+  for (const [key, value] of query) {
+    result[key] = value;
+  }
+  return result;
+}
```
JBWatenbergScality: I think query can be typed using URLSearchParams. Also I think this is equivalent to `Object.fromEntries(query.entries())`

src/react/databrowser/objects/ObjectDetails.jsx
```diff
@@ -67,21 +69,32 @@ function ObjectDetails({ toggled, listType }: Props) {
     }
     return null;
   };
-
+  function removeFirstTabFromSearchQuery(queryObject: TabQueryObject) {
+    delete queryObject.tab;
+    return queryObject;
+  }
```
JBWatenbergScality: Calling this function would mutate the original object introducing a side effect which may be difficult to debug if later on we want to access queryObject.tab. I suggest to refactor it to a pure function by making a deep copy of queryObject or by trying to pass {...queryObject, tab: null} to the summary tab.

src/react/databrowser/objects/MetadataSearch.jsx
```diff
@@ -58,11 +58,16 @@ const MetadataSearch = ({ isMetadataType, errorZenkoMsg }: Props) => {
...
   useOutsideClick(suggestionsRef, () => setHintsShown(false));
+  const searchInput = query.get('metadatasearch');
 
   const handleChange = e => {
     setInputText(e.target.value);
   };
 
+  useEffect(() => {
+    setInputText(searchInput);
+  }, []);
```
JBWatenbergScality: Instead of an effect I think this could be done when initiating input text state.
```js
const query = useQuery();
const searchInput = query.get('metadatasearch');
const [inputText, setInputText] = useState(searchInput || '');
```
That would avoid one useless rerender.

JBWatenbergScality (commented):
The build is failing due to 
`error  'useEffect' is defined but never used  no-unused-vars
 ` in src/react/databrowser/objects/MetadataSearch.jsx. Otherwise LGTM !

JBWatenbergScality (approved):

ChengYanJin:
Thanks for checking the behaviors on your side @nicolas2bert.  
I have checked with Valamir for this PR, and fixed some bugs.
Could you please take the time to review it again?

nicolas2bert (commented):
When I create a "folder" and enter it, the breadcrumb is not updated with the folder name.

ChengYanJin:
> When I create a "folder" and enter it, the breadcrumb is not updated with the folder name.

Hi Nicolas, 
Good catch!! I fixed in f81f53d55fe8c14587d64f04e02453bc19a84f85

src/react/ui-elements/Breadcrumb.jsx
```diff
@@ -28,18 +29,30 @@ const CustomBreadCrumb = styled(CoreUIBreadcrumb)`
...
+const breadcrumbPaths = (
+  pathname: string,
+  prefixPath: string,
+): Array<Element<'label'>> => {
   let match = matchPath(pathname, { path: '/create-bucket' });
   if (match) {
     return [<label key="buckets">create bucket</label>];
   }
-  match = matchPath(pathname, { path: '/buckets/:bucketName/objects/*' });
+  match = matchPath(pathname, { path: '/buckets/:bucketName/objects*' });
```
JBWatenbergScality: I would introduce a new var named `matchObjectRoutes` instead of mutating the `match` one

src/react/databrowser/objects/MetadataSearch.jsx
```diff
@@ -68,14 +70,18 @@ const MetadataSearch = ({
...
+    query.set('metadatasearch', inputText);
+    dispatch(push(`${pathname}?${query.toString()}`));
   };
 
   const reset = e => {
     e.stopPropagation();
     setHintsShown(false);
     setInputText('');
-    dispatch(listObjects(bucketName, prefixWithSlash));
+    // Remove the medatasearch from the query params
```
gaudiauj, line 82:
```suggestion
    // Remove the metadasearch from the query params
```

src/react/databrowser/objects/ObjectDetails.jsx
```diff
@@ -71,17 +71,25 @@ function ObjectDetails({ toggled, listType }: Props) {
   return (
     <ContentSection>
       <CustomTabs>
-        <CustomTabs.Tab label="Summary" path={pathname}>
+        <CustomTabs.Tab
+          label="Summary"
+          path={pathname}
+          query={{ ...queryObject, ...{ tab: null } }}
```
gaudiauj, line 77:
```suggestion
          query={{ ...queryObject, tab: null }}
```

src/react/databrowser/objects/ObjectDetails.jsx
```diff
@@ -71,17 +71,25 @@ function ObjectDetails({ toggled, listType }: Props) {
...
+        >
           {details()}
         </CustomTabs.Tab>
         <CustomTabs.Tab
           label="Metadata"
           path={pathname}
-          query={{ tab: 'metadata' }}
+          query={{ ...queryObject, ...{ tab: 'metadata' } }}
```
gaudiauj, line 84:
```suggestion
          query={{ ...queryObject, tab: 'metadata'  }}
```

src/react/databrowser/objects/ObjectDetails.jsx
```diff
@@ -71,17 +71,25 @@ function ObjectDetails({ toggled, listType }: Props) {
...
         >
           {details()}
         </CustomTabs.Tab>
-        <CustomTabs.Tab label="Tags" path={pathname} query={{ tab: 'tags' }}>
+        <CustomTabs.Tab
+          label="Tags"
+          path={pathname}
+          query={{ ...queryObject, ...{ tab: 'tags' } }}
```
gaudiauj, line 91:
```suggestion
          query={{ ...queryObject, tab: 'tags' }}
```

src/react/databrowser/objects/ObjectListTable.jsx
```diff
@@ -73,11 +74,42 @@ export default function ObjectListTable({
   const handleCellClicked = useCallback(
     (bucketName, key) => e => {
       e.stopPropagation();
-      dispatch(push(`/buckets/${bucketName}/objects/${key}`));
+      query.set('prefix', key);
+      dispatch(push(`/buckets/${bucketName}/objects?${query.toString()}`));
     },
     [dispatch],
```
gaudiauj, line 80: just for eslint to be happy, it shouldn't change the code beahavior
```suggestion
    [dispatch, query],
```

src/react/databrowser/objects/ObjectRow.jsx
```diff
@@ -58,14 +62,33 @@ const Row = ({
...
+    dispatch(push(`${pathname}?${query.toString()}`));
   };
-
+  const isRowSelected = isListVersions
+    ? row.original.versionId === versionId && objectKey === row.original.key
+    : objectKey === row.original.key;
   return (
     <T.Row
-      isSelected={row.original.toggled}
+      isSelected={row.original.toggled || isRowSelected}
```
gaudiauj, line 91:
```suggestion
      isSelected={isRowSelected}
```

src/react/databrowser/objects/ObjectRow.jsx
```diff
@@ -58,14 +62,33 @@ const Row = ({
...
+      query.set('versionId', row.original.versionId);
+    } else {
+      query.delete('versionId');
+    }
+    dispatch(push(`${pathname}?${query.toString()}`));
   };
-
+  const isRowSelected = isListVersions
+    ? row.original.versionId === versionId && objectKey === row.original.key
+    : objectKey === row.original.key;
```
gaudiauj, line 88:
```suggestion
  const isRowSelected = (isListVersions
    ? row.original.versionId === versionId && objectKey === row.original.key
    : objectKey === row.original.key) || row.original.toggled;
```

JBWatenbergScality (approved):

src/react/databrowser/objects/Objects.jsx
```diff
@@ -30,14 +33,40 @@ export default function Objects() {
...
+        setLoaded(true),
+      );
+    } else {
+      dispatch(
+        listObjects(
+          bucketNameParam,
+          prefixWithSlash,
+          isShowVersions ? LIST_OBJECT_VERSIONS_S3_TYPE : LIST_OBJECTS_S3_TYPE,
+        ),
+      ).finally(() => setLoaded(true));
```
gaudiauj, line 67: i'm not sure the dispatch function work as a async function. Not sure that the finally will run after all actions are done.

gaudiauj (approved):
just few comment about small code improvements. Otherwise look nice !

src/react/databrowser/objects/Objects.jsx
ChengYanJin: Since here I only have one action, it should be OK.

PR #216, opened 2022-01-05
ZKUI-58 Enabling Async Notification should enable versioning

nicolas2bert:
Screen-Recording-2022-01-05-at-4 00 47-PM

### DIFF
--- a/src/react/databrowser/buckets/BucketCreate.jsx
+++ b/src/react/databrowser/buckets/BucketCreate.jsx
@@ -1,12 +1,12 @@
 // @flow
 import { Banner, Toggle } from '@scality/core-ui';
-import { SmallerText } from '@scality/core-ui/dist/components/text/Text.component';
 import SpacedBox from '@scality/core-ui/dist/components/spacedbox/SpacedBox';
 import { Controller, FormProvider, useForm } from 'react-hook-form';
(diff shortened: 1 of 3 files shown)


naren-rajendran:
@nicolas2bert I am not sure if it is related..but why is `versioning` either `active` or `inactive` and `async notifications` has `enabled` or `disabled` shouldn't they be the same? is there a difference or reasoning behind this?

nicolas2bert:
Thanks! It is not directly related to this PR but good catch. I created a ticket: https://scality.atlassian.net/browse/ZKUI-60 and will double check with @Cuervino to make sure it is the expected wording.

src/react/databrowser/buckets/BucketCreate.jsx
```diff
@@ -153,6 +154,12 @@ function BucketCreate() {
     return `${option.name} (${locationTypeName})`;
   };
 
+  const matchVersionning = (checked: boolean) => {
```
miniscruff:
```suggestion
  const matchVersioning = (checked: boolean) => {
```
One too many N's

miniscruff (approved):

nicolas2bert:
@bert-e approve

nicolas2bert:
ping

naren-rajendran (approved):

PR #226, opened 2022-01-21
ZKUI-80: Add missing header when deleting object

ChengYanJin:
1. Add the missing header when deleting an object.
2. Display the modal with an error message when deleting object with COMPLIANCE mode 
   S3 API returns **Access Denied** but with 200 response code. 
   
image

3. Fix the small bug that we have to increase retention period when switching form GOVERNANCE mode to COMPLIANCE mode

### DIFF
--- a/src/js/S3Client.js
+++ b/src/js/S3Client.js
@@ -237,9 +237,18 @@ export default class S3Client {
         Delete: {
           Objects: objects,
         },
+        BypassGovernanceRetention: true,
       };
       return this.client.deleteObjects(params, (error, data) => {
-        error ? reject(error) : resolve(data);
+        const hasAccessDeniedError =
+          data &&
+          data.Errors &&
+          data.Errors.some(error => error.Code === 'AccessDenied');
+        if (error || hasAccessDeniedError) {
+          reject(error || data);
+        }
+
+        resolve(data);
       });
     }).then(() => this._deleteFolders(bucketName, folders));
   }
--- a/src/react/actions/s3object.js
+++ b/src/react/actions/s3object.js
@@ -31,7 +31,11 @@ import type {
   S3Version,
   TagSet,
 } from '../../types/s3';
-import { handleAWSClientError, handleAWSError } from './error';
+import {
+  handleAWSClientError,
+  handleAWSError,
+  handleErrorMessage,
+} from './error';
 import { networkEnd, networkStart } from './network';
 import { LIST_OBJECT_VERSIONS_S3_TYPE } from '../utils/s3';
 import type { Marker, ZenkoClient } from '../../types/zenko';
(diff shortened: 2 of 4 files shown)


src/react/actions/s3object.js
```diff
@@ -467,7 +471,20 @@ export function deleteFiles(
...
       .deleteObjects(bucketName, objects, folders)
-      .catch(error => dispatch(handleAWSClientError(error)))
+      .then(data => {
+        // S3 API return 200 response code with Access Denied Error
+        const hasAccessDeniedError =
+          data &&
+          data.Errors &&
+          data.Errors.some(error => error.Code === 'AccessDenied');
+        if (hasAccessDeniedError) {
+          const str = 'You do not have the permission to this action.';
```
JBWatenbergScality:
```suggestion
          const str = 'You do not have the permission to perform this action.';
```

src/js/S3Client.js
```diff
@@ -237,11 +237,12 @@ export default class S3Client {
         Delete: {
           Objects: objects,
         },
+        BypassGovernanceRetention: true,
       };
       return this.client.deleteObjects(params, (error, data) => {
         error ? reject(error) : resolve(data);
       });
-    }).then(() => this._deleteFolders(bucketName, folders));
+    }).then(data => this._deleteFolders(bucketName, folders, data));
```
JBWatenbergScality: What if we try to delete some folders ? The current implementation wouldn't raise any error I think.
I would go for a different approach where the error handling is performed in the client wrapper instead of delegating it to the client consumer. 

```suggestion

        BypassGovernanceRetention: true,
      };
      return this.client.deleteObjects(params, (error, data) => {
        const hasAccessDeniedError =
          data &&
          data.Errors &&
          data.Errors.some(error => error.Code === 'AccessDenied');

        if (error || hasAccessDeniedError){
           reject(error || data);
        }
        
        resolve(data);
      });
    }).then(() => this._deleteFolders(bucketName, folders));
    }).then(() => this._deleteFolders(bucketName, folders));
```

And then we keep the old  _deleteFolders  implementation.

src/react/actions/s3object.js
```diff
@@ -467,7 +471,20 @@ export function deleteFiles(
...
+          dispatch(handleErrorMessage(str, 'byModal'));
+        }
```
JBWatenbergScality: With my above suggestion, this should now be handled by the `catch` clause.

JBWatenbergScality (approved):

src/js/S3Client.js
```diff
@@ -237,9 +237,18 @@ export default class S3Client {
...
           Objects: objects,
         },
+        BypassGovernanceRetention: true,
       };
       return this.client.deleteObjects(params, (error, data) => {
-        error ? reject(error) : resolve(data);
+        const hasAccessDeniedError =
+          data &&
+          data.Errors &&
+          data.Errors.some(error => error.Code === 'AccessDenied');
```
gaudiauj, line 246: other error should be ignored here ?

gaudiauj (approved):
looks good to me

src/js/S3Client.js
ChengYanJin: if we have other errors we reject them directly. 
the problem with this AccessDenied is we get 200 response code from the API, so we have to check explicitly if we have this error in the data field.

PR #229, opened 2022-01-25
Zk UI 82 button to delete workflow closes itself instantly

gaudiauj:
fix bug when clicking delete workflow.

since the button type was not specified, clicking on the button would validate the edit form workflow which would reload the page.

### DIFF
--- a/src/react/workflow/details/Configuration.jsx
+++ b/src/react/workflow/details/Configuration.jsx
@@ -85,6 +85,7 @@ function Configuration({
               label="Delete Workflow"
               variant="danger"
               onClick={handleOpenDeleteModal}
+              type="button"
             />
           </T.Header>
           <Replication


src/react/workflow/details/Configuration.jsx
```diff
@@ -85,6 +85,7 @@ function Configuration({
               label="Delete Workflow"
               variant="danger"
               onClick={handleOpenDeleteModal}
+              type="button"
```
ChengYanJin, line 88: should we add this type in the core-ui then?

gaudiauj: it's from the HTML specification. Default type is submit. Not sure it's needed in core-ui since it's classic HTML

ChengYanJin (approved):

MonPote (approved):

ChengYanJin:
close the pull request because of the wrong branch name.

PR #233, opened 2022-01-31
ZKUI-86 : add error message on bucket creation

gaudiauj:
add error message and ui validation on bucket creation : `Bucket names can include only lowercase letters, numbers, dots (.), and hyphens (-)`

### DIFF
--- a/src/react/databrowser/buckets/BucketCreate.jsx
+++ b/src/react/databrowser/buckets/BucketCreate.jsx
@@ -25,11 +25,15 @@ import ObjectLockRetentionSettings, {
 } from './ObjectLockRetentionSettings';
 import { XDM_FEATURE } from '../../../js/config';
 
+export const bucketErrorMessage =
+  'Bucket names can include only lowercase letters, numbers, dots (.), and hyphens (-)';
+
 const schema = Joi.object({
   name: Joi.string()
     .label('Name')
     .required()
     .min(3)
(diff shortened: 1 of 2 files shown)


src/react/databrowser/buckets/BucketCreate.jsx
```diff
@@ -174,8 +178,9 @@ function BucketCreate() {
                 tooltipMessages={[
                   'Must be unique',
                   'Cannot be modified after creation',
+                  bucketErrorMessage,
```
ChengYanJin, line 181: I think we were saying we don't need to add this information to the tooltip, only the dynamic validation error is enough.
Can you double-check with @Cuervino ?

gaudiauj: he commented on the ticket : `This text will also be added in the tooltip next to “Bucket name”.`
but i'll ask him anyway

gaudiauj: he confirmed this has to go on the tooltip as well :)

ChengYanJin (approved):
LGTM!

src/react/databrowser/buckets/BucketCreate.jsx
ChengYanJin: OK. sorry! I didn't really check the comments on the tickets. Thanks @Cuervino !

Cuervino (approved):
LGTM

PR #235, opened 2022-02-07
ZKUI-87: Create account list screen

ChengYanJin:
- Separate the account list with the account tab
- Update the account list with the new table component
- Remove the `Accounts.test` and `AccountList.test` (will implement the test in the coming PR)
- we are missing the link component in core-UI. it will come just after

image

image


JBWatenbergScality (commented):
ok for me at this stage

MonPote (approved):
Let's go

JBWatenbergScality (approved):

PR #238, opened 2022-02-10
ZKUI-72: Listing IAM users and access keys

MonPote:
List IAM USER
image

List Access key 
image

not included in this PR:
- testing 
- handle the loading  status of the table

### DIFF
--- a/src/react/IAMProvider.jsx
+++ b/src/react/IAMProvider.jsx
@@ -1,0 +1,45 @@
+import React, { createContext, useContext, useState } from 'react';
+import { useSelector } from 'react-redux';
+import { useParams } from 'react-router-dom';
+import { getAssumeRoleWithWebIdentityIAM } from '../js/IAMClient';
+import { useQuery } from 'react-query';
+
+const IAMContext = createContext(null);
+
+export const useIAMClient = () => {
+  const IAMCtxt = useContext(IAMContext);
+
+  if (!IAMCtxt) {
+    throw new Error(
+      'The useIAMClient hook can only be used within IAMProvider.',
+    );
+  }
+
+  return IAMCtxt.iamClient;
+};
+
+const IAMProvider = ({ children }) => {
+  const { accountName } = useParams();
+  const [IAMClient, setIAMClient] = useState(null);
+  const state = useSelector(state => state);
+
+  useQuery({
+    queryKey: ['updateIAMClient', accountName],
+    queryFn: () => {
+      return getAssumeRoleWithWebIdentityIAM(state, accountName).then(
+        iamClient => {
+          setIAMClient(iamClient);
+        },
+      );
+    },
+    enabled: accountName && accountName !== '',
+  });
+
+  return (
+    <IAMContext.Provider value={{ iamClient: IAMClient }}>
+      {children}
+    </IAMContext.Provider>
+  );
+};
+
+export default IAMProvider;
--- a/src/react/IAMProvider.jsx
+++ b/src/react/IAMProvider.jsx
@@ -1,10 +1,12 @@
-import React, { createContext, useContext, useState } from 'react';
+//@flow
+import React, { createContext, useContext } from 'react';
+import type { Node } from 'react';
 import { useSelector } from 'react-redux';
 import { useParams } from 'react-router-dom';
-import { getAssumeRoleWithWebIdentityIAM } from '../js/IAMClient';
+import IAMClient, { getAssumeRoleWithWebIdentityIAM } from '../js/IAMClient';
 import { useQuery } from 'react-query';
 
-const IAMContext = createContext(null);
+const IAMContext = createContext<null | IAMClient>(null);
 
 export const useIAMClient = () => {
   const IAMCtxt = useContext(IAMContext);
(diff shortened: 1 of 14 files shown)


src/react/IAMProvider.jsx
```diff
@@ -0,0 +1,45 @@
...
+
+  const IAMClient = useMemo(() => {
+    if (status === 'success') {
+      return IAMClientResult;
+    }
+    return null;
+  }, [status]);
+
+  return (
+    <IAMContext.Provider value={{ iamClient: IAMClient }}>
```
JBWatenbergScality:
```suggestion
  return (
    <IAMContext.Provider value={{ iamClient: IAMClientResult || null }}>
```

src/react/account/AccountCreateUser.jsx
```diff
@@ -0,0 +1,152 @@
...
+  const IAMClient = useIAMClient();
+  const { register, handleSubmit, errors } = useForm({
+    resolver: joiResolver(schema),
+  });
+
+  const createAccessKeyMutation = useMutation(userName => {
+    dispatch(networkStart('Creating User'));
+    return IAMClient.createUser(userName)
+      .then(() => {
+        queryClient.invalidateQueries('listIAMClient');
```
JBWatenbergScality: A quick search on `listIAMClient` is not raising any matches, we probably want to invalidate `['listIAMUsers', accountName]`

src/react/account/AccountCreateUser.jsx
```diff
@@ -0,0 +1,152 @@
...
+        // invalidate stuff
+        // handle error ?
+      })
```
JBWatenbergScality, line 54: In order to handle the error we can just dispatch the existing error actions for now I think 🤔

src/react/account/AccountUserAccessKeys.jsx
```diff
@@ -0,0 +1,137 @@
...
+    },
+    enabled: IAMClient !== null,
+  });
+
+  const createdOnCell = ({ value }) => {
+    return (
+      <div>
+        {isAccessKeyOutDated(value) ? (
+          <Icon name={'Exclamation-circle'} color={'statusWarning'} />
+        ) : null}
```
JBWatenbergScality: The user won't know why there is such a waring next to the creation date. I think likely when hovering it we should get an explanation tooltip @Cuervino WDYT ?
Additionally the "warning" computation should be abstracted away so that we can easily replace it with an alert when alerting will be configured. I would take the approach to implement a custom hook `const outdatedAlert = useAccessKeyOutdatedStatus(theWholeAccessKeyObject)` which for now would just check creation time but later on could easily be refactored to alert checking. `outdatedAlert` would be either null either `This access key has been generated more than 300 days ago. Please consider change it.`, text to be validated of course.

src/react/account/AccountUserList.jsx
```diff
@@ -0,0 +1,160 @@
...
+      return 0;
+    }, [userAccessKeyStatus]);
+
+    return (
+      <div>
+        {accessKeys >= 2 ? (
+          <TextBadge variant={'statusWarning'} text={accessKeys}></TextBadge>
+        ) : (
+          accessKeys
+        )}
```
JBWatenbergScality: We are missing error and loading cases here

src/react/account/AccountUserList.jsx
```diff
@@ -0,0 +1,160 @@
...
+            overlayStyle: { width: '10rem' },
+            overlay: 'Checking the access keys',
+          }}
+        />
+      </div>
+    );
+  };
+  const renderAccessKeyComponent = ({ row }) => (
+    <AsyncRenderAccessKey rowValues={row.original} />
+  );
```
JBWatenbergScality: I think it would be better to extract those components outside of the AccountUserList definition, you can define them just on top of this file.

src/react/account/AccountUserSecretKeyModal.jsx
```diff
@@ -0,0 +1,136 @@
...
+
+  const createAccessKeyMutation = useMutation(userName => {
+    return IAMClient.createAccessKey(userName).then(res => {
+      const newKey = {
+        accountName: res.AccessKey.UserName,
+        accessKey: res.AccessKey.AccessKeyId,
+        secretKey: res.AccessKey.SecretAccessKey,
+      };
+      setNewKey(newKey);
+      queryClient.invalidateQueries('listIAMClientUserAccessKeys');
```
JBWatenbergScality: I don't think this is the right invalidated query

src/react/account/AccountUserAccessKeys.jsx
```diff
@@ -0,0 +1,137 @@
...
+
+const AccountUserAccessKeys = () => {
+  const { pathname } = useLocation();
+  const { IAMUserName } = useParams();
+  const history = useHistory();
+  const IAMClient = useIAMClient();
+  const { url } = useRouteMatch();
+
+  const accessKeysQuery = useQuery({
+    queryKey: ['listIAMClientUserAccessKeys', IAMUserName],
```
JBWatenbergScality: It likely should be the same query as the one used in the table to share cache keys ([line 56 in src/react/account/AccountUserList.jsx](https://github.com/scality/zenko-ui/pull/238/files#diff-36a1ebb41c89476d16eaee94f9f30a23f5f8f3d0bc7032d1030def24904d4a3aR55-R59) )

JBWatenbergScality (changes requested):

JBWatenbergScality (approved):

ChengYanJin (approved):

PR #246, opened 2022-02-24
ZKUI-97: Add copy arn button in the user list

ChengYanJin:
image

### DIFF
--- a/src/react/ui-elements/CopyARNButton.jsx
+++ b/src/react/ui-elements/CopyARNButton.jsx
@@ -1,0 +1,38 @@
+//@flow
+import React from 'react';
+import styled from 'styled-components';
+import { Button } from '@scality/core-ui/dist/next';
+import { spacing } from '@scality/core-ui/dist/style/theme';
+import { COPY_STATE_SUCCESS, useClipboard } from '../utils/hooks';
+
+const InlineButton = styled(Button)`
+  height: ${spacing.sp24};
+  margin-left: ${spacing.sp16};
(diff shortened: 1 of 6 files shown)


src/react/ui-elements/CopyARNButton.jsx
```diff
@@ -0,0 +1,55 @@
...
+  }, [copySuccess]);
+
+  const copyToClipboard = () => {
+    if (!navigator || !navigator.clipboard) {
+      return;
+    }
+    navigator.clipboard.writeText(text);
+    setCopySuccess(true);
+  };
+
```
JBWatenbergScality: The logic seems to be duplicated with the one of the Clipboard component. I vote for a refactoring of the existing component such that:
 - we create a custom hook `useClipboard` defined as :
```
type useClipboard = () => {copy: (text: string) => void, status: 'idle' | 'success' | 'unsupported'} 
``` 
 - we create a generic component with a children props wrapping this children with the tooltip mentioning the success of the copy
 - we change the existing clipboard component to use this generic one
 - we change this implementation with the generic component
-

JBWatenbergScality (changes requested):

ChengYanJin:
Hello @Cuervino
 
We suggest removing the **tooltip** "Copied!" which provides the feedback of the copy action.

Instead of this behavior,
We will temporarily change the text of the Copy button to"Copied" with a green tick icon.

WDYT?

JBWatenbergScality (approved):

MonPote (approved):

PR #265, opened 2022-03-07
Edit users

AsmaaELM:
Issue: https://scality.atlassian.net/browse/ZKUI-96

### DIFF
--- a/src/react/account/AccountUpdateUser.tsx
+++ b/src/react/account/AccountUpdateUser.tsx
@@ -1,0 +1,192 @@
+import FormContainer, * as F from '../ui-elements/FormLayout';
+import React, { useRef } from 'react';
+import { clearError, handleErrorMessage, networkEnd } from '../actions';
+import { useDispatch, useSelector } from 'react-redux';
+import type { AppState } from '../../types/state';
+import { Banner } from '@scality/core-ui';
+import { Button } from '@scality/core-ui/dist/next';
+import Joi from '@hapi/joi';
+import { joiResolver } from '@hookform/resolvers';
+import { useForm } from 'react-hook-form';
+import { useOutsideClick } from '../utils/hooks';
+import { useIAMClient } from '../IAMProvider';
+import { queryClient } from '../App';
+import { InfiniteData, useMutation } from 'react-query';
+import { useHistory, useParams } from 'react-router-dom';
+import { notFalsyTypeGuard } from '../../types/typeGuards';
+import { ListUsersResponse } from 'aws-sdk/clients/iam';
+const regexpName = /^[\w+=,.@ -]+$/;
+const schema = Joi.object({
+  name: Joi.string()
+    .label('Name')
+    .required()
+    .min(2)
+    .max(64)
+    .regex(regexpName)
+    .message('Invalid Name'),
+});
+
+const AccountUpdateUser = () => {
+  const dispatch = useDispatch();
+  const history = useHistory();
+  const IAMClient = useIAMClient();
+  const { IAMUserName, accountName } = useParams<{ IAMUserName: string, accountName: string }>();
+  const { register, handleSubmit, errors } = useForm({
+    resolver: joiResolver(schema)
+  });
+  const updateUserMutation = useMutation(
+    (newUserName: string) => {
+      const oldUserName = IAMUserName;
+      return notFalsyTypeGuard(IAMClient).updateUser(newUserName, oldUserName)
+        .then(() => {
+          const olddata = queryClient.getQueryData([
+            'listIAMUsers',
+            accountName,
+          ]);
+          olddata && queryClient.setQueryData<InfiniteData<ListUsersResponse>>(
+            ['listIAMUsers', accountName], (old) => {
+              const pages = notFalsyTypeGuard(old).pages.map((page) => {
+                const users = page.Users;
+                const index = users.findIndex(
+                  (user) => user.UserName === oldUserName,
+                );
+                if(index !== -1) {
+                  users[index].UserName = newUserName;
+                }
+                return page;
+              });
+              return {
+                ...notFalsyTypeGuard(old),
+                pages,
+              };
+          });
+        })
+        .catch((err) => {
+          dispatch(handleErrorMessage(`${err}`, 'byModal'));
+        })
+        .finally(() => {
+          dispatch(networkEnd());
+        });
+    },
+    {
+      onSuccess: () => {
+        history.goBack();
+      }
+    },
+  );
+
+  /**
+   * This part has to be handle
+   */
+  const hasError = useSelector(
+    (state: AppState) =>
+      !!state.uiErrors.errorMsg && state.uiErrors.errorType === 'byComponent',
+  );
+  const errorMessage = useSelector(
+    (state: AppState) => state.uiErrors.errorMsg,
+  );
+  const loading = useSelector(
+    (state: AppState) => state.networkActivity.counter > 0,
+  );
+
+  const onSubmit = ({ name }: {name: string}) => {
+    clearServerError();
+    updateUserMutation.mutate(name);
+  };
+
+  const handleCancel = (e: MouseEvent) => {
+    if (e) {
+      e.preventDefault();
+    }
+
+    clearServerError();
+    history.goBack();
+  };
+
+  const clearServerError = () => {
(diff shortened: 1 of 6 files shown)


src/react/account/AccountUpdateUser.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+const AccountUpdateUser = () => {
+  const dispatch = useDispatch();
+  const history = useHistory();
+  const IAMClient = useIAMClient();
+  const { IAMUserName } = useParams();
+  const { register, handleSubmit, errors } = useForm({
+    resolver: joiResolver(schema),
+  });
+  const { accountName } = useParams();
```
MonPote: I think you have something like : 
```suggestion
  const { IAMUserName, accountName } = useParams<{ IAMUserName: string, accountName: string }>();
  const { register, handleSubmit, errors } = useForm({
    resolver: joiResolver(schema),
  });
```

src/react/account/AccountUpdateUser.tsx
```diff
@@ -0,0 +1,191 @@
...
+  const loading = useSelector(
+    (state: AppState) => state.networkActivity.counter > 0,
+  );
+
+  const onSubmit = ({ name }: {name: string}) => {
+    clearServerError();
+    updateUserMutation.mutate(name);
+  };
+
+  const handleCancel = (e: any) => {
```
MonPote:
```suggestion
  const handleCancel = (e: MouseEvent<HTMLButtonElement>) => {
```

So you will also need to add `MouseEvent` in React import.

src/react/account/AccountUpdateUser.tsx
```diff
@@ -0,0 +1,191 @@
...
+  const { IAMUserName, accountName } = useParams<{ IAMUserName: string, accountName: string }>();
+  const { register, handleSubmit, errors } = useForm({
+    resolver: joiResolver(schema)
+  });
+  const updateUserMutation = useMutation(
+    (newUserName: string) => {
+      const oldUserName = IAMUserName;
+      return notFalsyTypeGuard(IAMClient).updateUser(newUserName, oldUserName)
+        .then(() => {
+          queryClient.setQueryData(['listIAMUsers', accountName], (old: any) => {
```
MonPote: You should type queryClient.setQueryData like this : 
This way, you won't need to type anything in your function, everything will be infer by Typescript.
```suggestion
            queryClient.setQueryData<InfiniteData<ListUsersResponse>>(
              ['listIAMUsers', accountName],
              (old) => {
```

src/react/account/AccountUpdateUser.tsx
```diff
@@ -0,0 +1,191 @@
...
+            if (old) {
+              const pages: any = old.pages;
+              pages.some((page: any) => {
```
MonPote: I think you should use `map` instead of some here.

src/react/account/AccountUpdateUser.tsx
```diff
@@ -0,0 +1,191 @@
...
+                const users = page.Users;
+                const index = users.findIndex(
+                  (user: any) => user.UserName === oldUserName,
+                );
+                return index !== -1 && (users[index].UserName = newUserName);
```
MonPote: You should not assign a value in a return, it is : 
1. hard to read because and reviewer will need to process in they mind in order to understand this.
2. -> hard de debug something goes wrong 

When you get refactor with map, please make sure to at least assign the value in a variable before returning it.

src/react/account/AccountUpdateUser.tsx
```diff
@@ -0,0 +1,191 @@
...
+  const IAMClient = useIAMClient();
+  const { IAMUserName, accountName } = useParams<{ IAMUserName: string, accountName: string }>();
+  const { register, handleSubmit, errors } = useForm({
+    resolver: joiResolver(schema)
+  });
+  const updateUserMutation = useMutation(
+    (newUserName: string) => {
+      const oldUserName = IAMUserName;
+      return notFalsyTypeGuard(IAMClient).updateUser(newUserName, oldUserName)
+        .then(() => {
```
MonPote, line 41: Right now when you arrive directly on the page and try to edit the name it will crash.
In order to handle this "edge cases" you should check if you have data first.

In the `then`, you should first check : 
```suggestion
        .then(() => {
          const olddata = queryClient.getQueryData([
            'listIAMUsers',
            accountName,
          ]);
          if (olddata) {
             queryClient.setQueryData...
             [...]
          }
```

src/react/account/AccountUpdateUser.tsx
```diff
@@ -0,0 +1,191 @@
...
+            User Name
+          </F.Label>
+          <F.Input
+            type="text"
+            id="name"
+            name="name"
+            ref={register}
+            onChange={clearServerError}
+            autoComplete="new-password"
+          />
```
MonPote, line 150: The behaviour is not correct, we currently have this : 


instead of this like in the create user : 


You should fix the error position and the label position.
You should also user `Name` instead of `User Name` for more consistency.

src/react/account/AccountUpdateUser.tsx
```diff
@@ -0,0 +1,191 @@
...
+              <div>The User stays in the same Groups, under its new name.</div>
+              <br />
+              <div>Policies:</div>
+              <div>- Any Policies attached to the user stays with this user, under its new name.</div>
+              <div>- Any Role (Trust) Policies that refer to the User as a Principal are automatically updated with the
+                new name.
+              </div>
+              <div>- Any Policies that refer to the User as a Resource are not updated, you have to do it manually.
+              </div>
+            </div>).toString()]
```
MonPote: I believe you add `.toString()` to handle the Typescript errors.
However you now have this : 


The correct way to fix Typescript Errors should be check what type is missing.
In this case, it's from the `LabelProps` (src/react/ui-elements/FormLayout.tsx)

```diff
type LabelProps = {
  children: JSX.Element | string;
-  tooltipMessages?: Array<string>;
+  tooltipMessages?: Array<JSX.Element | string>;
  tooltipWidth?: string;
  style?: CSSStyleSheet;
};
```

src/react/ui-elements/FormLayout.tsx
```diff
@@ -45,7 +45,7 @@ export const SubTitle = styled.div`
 `;
 export const Fieldset = styled.fieldset`
   display: flex;
-  flex-direction: ${(props) => props.direction || 'column'};
+  flex-direction: ${props => props.flexDirection || 'column'};
```
MonPote: This change is useless, because you are using the props.style cf : 
```js
<F.Fieldset style={{ width: '100%', flexDirection: 'row' }}>
```

If you want really want to do this change you should have something like this : 
```js
<F.Fieldset flexDirection style={{ width: '100%' }}>
```

src/react/ui-elements/FormLayout.tsx
```diff
@@ -89,8 +89,9 @@ const WarningInputContainer = styled.div`
   color: ${(props) => props.theme.brand.warning};
 `;
 type ErrorInputProps = {
-  children: JSX.Element;
+  children: JSX.Element|JSX.Element[];
```
MonPote, line 92: Good 👍

src/react/ui-elements/FormLayout.tsx
```diff
@@ -115,7 +116,7 @@ const UlOverlay = styled.ul`
   padding: 0px 0px 0px ${spacing.sp20};
 `;
 type LabelProps = {
-  children: JSX.Element;
+  children: JSX.Element|string;
   tooltipMessages?: Array<string>;
```
MonPote: Already asked the change in previous file

MonPote (changes requested):

src/react/account/AccountUpdateUser.tsx
Cuervino: @MonPote @AsmaaELM 
Indeed, the figma screens lacked consistency.
The layout should be "Top", and not "Side", in this case.
I amended the screens.
image

src/react/ui-elements/FormLayout.tsx
AsmaaELM: Thanks!

AsmaaELM:
ping

MonPote (approved):

PR #271, opened 2022-03-11
Testing account users

AsmaaELM:
Issue: https://scality.atlassian.net/browse/ZKUI-69

### DIFF
--- a/src/react/account/AccountUserList.tsx
+++ b/src/react/account/AccountUserList.tsx
@@ -1,4 +1,4 @@
-import { ChangeEvent, useMemo } from 'react';
+import React, { ChangeEvent, useMemo } from 'react';
 import styled from 'styled-components';
 import { useHistory, useRouteMatch } from 'react-router-dom';
 import { Table, Button } from '@scality/core-ui/dist/next';
@@ -92,9 +92,28 @@ const renderAccessKeyComponent = ({ row }) => (
   <AsyncRenderAccessKey userName={row.original.userName} />
 );
 
+const RenderEditButton = ({ userName }: { userName: string }) => {
+  const history = useHistory();
+  return (
+    <SpacedBox ml={12}>
+      <Button
+        style={{ height: spacing.sp24 }}
+        variant='secondary'
+        label='Edit'
+        icon={<i className='fa fa-pen'></i>}
+        onClick={() => history.push(`users/${userName}/update-user`)}
+      />
+    </SpacedBox>);
+};
+
 const renderActionButtons = (rowValues) => {
-  const { arn } = rowValues;
-  return <CopyARNButton text={arn} />;
+  const { arn, userName } = rowValues;
+  return (
+    <div style={{ display: 'flex' }}>
+      <CopyARNButton text={arn} />
+      <RenderEditButton userName={userName} />
+    </div>
+  );
 };
 
 const WithTooltipWhileLoading = ({
@@ -199,6 +218,7 @@ const AccountUserList = ({ accountName }: { accountName?: string }) => {
       accessor: 'actions',
       cellStyle: {
         textAlign: 'right',
+        marginLeft: 'auto',
         minWidth: '25rem',
       },
       disableSortBy: true,
--- a/src/react/account/__tests__/AccountUserList.test.tsx
+++ b/src/react/account/__tests__/AccountUserList.test.tsx
@@ -1,19 +1,22 @@
-import { render, screen, waitFor, fireEvent } from '@testing-library/react';
+import {
+  render,
+  screen,
+  waitFor,
+  fireEvent,
+  getAllByRole,
+  getByText
+} from '@testing-library/react';
 import '@testing-library/jest-dom';
-import { createMemoryHistory } from 'history';
-import { Router } from 'react-router-dom';
-import IAMClient from '../../../js/IAMClient';
-import { _IAMContext } from '../../IAMProvider';
 import AccountUserList from '../AccountUserList';
-import { QueryClient, QueryClientProvider } from 'react-query';
 import { rest } from 'msw';
 import { setupServer } from 'msw/node';
-import { ReactNode } from 'react';
+import { wrapper } from '../../utils/test';
 
 const SAMPLE_USER_ID = 'GENERATED_ID';
 const SAMPLE_USER_NAME = 'test';
 const SAMPLE_CREATE_DATE = '2022-03-02T08:35:24Z';
 const SAMPLE_ARN = `arn:aws:iam::970343539682:user/${SAMPLE_USER_NAME}`;
+const nbrOfColumnsExpectedWithoutEdit = 4;
 
 // AutoSizer uses offsetWidth and offsetHeight.
 // Jest runs in JSDom which doesn't support measurements APIs.
@@ -25,7 +28,7 @@ function mockOffsetSize(width: number, height: number) {
         paddingLeft: 10,
         paddingRight: 10,
         paddingBottom: 10,
-        fontSize: 14
+        fontSize: 14,
       };
     },
   });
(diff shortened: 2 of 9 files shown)


src/react/account/AccountUserList.tsx
```diff
@@ -140,6 +140,7 @@ const AccountUserList = ({ accountName }: { accountName?: string }) => {
         return IAMClient.listUsers(1000, marker);
       },
       staleTime: Infinity,
+      onError: console.log,
```
JBWatenbergScality:
```suggestion
```

src/react/account/__tests__/AccountUserList.test.tsx
```diff
@@ -1,4 +1,12 @@
-import { render, screen, waitFor, fireEvent } from '@testing-library/react';
+import {
+  render,
+  screen,
+  waitFor,
+  fireEvent,
+  prettyDOM,
```
JBWatenbergScality: Please take care of github action warning below:
```
src/react/account/__tests__/AccountUserList.test.tsx#L6
'prettyDOM' is defined but never used
```

src/react/account/__tests__/AccountUserList.test.tsx
```diff
@@ -88,7 +97,14 @@ const wrapper = ({ children }: { children: ReactNode}) => {
   const iamClient = new IAMClient('http://testendpoint');
   iamClient.login(params);
   return (
-    <QueryClientProvider client={new QueryClient()}>
+    <QueryClientProvider client={new QueryClient({
```
JBWatenbergScality: The wrapper should likely be moved to a common utils for test file to be mutualised with other tests.
Actually it is already present in `src/react/utils/test.tsx` bellow, so you would just have to factorize this change there and drop this one.

src/react/account/__tests__/AccountUserList.test.tsx
```diff
@@ -124,5 +140,88 @@ describe('AccountUserList', () => {
     expect(screen.getByText(SAMPLE_USER_NAME)).toBeInTheDocument();
 
     //TODO test copy arn
```
JBWatenbergScality: This PR address this todo hence it can be removed
```suggestion
```

src/react/account/__tests__/AccountUserList.test.tsx
```diff
@@ -124,5 +140,88 @@ describe('AccountUserList', () => {
...
+    expect(getByText(
+      arnButtonOfFirstRow[1],
+      /Copy ARN/i,
+    )).toBeInTheDocument();
+
+    fireEvent.mouseEnter(screen.getAllByRole('button')[4]);
+
+    expect(
+      screen.getByText('Copy ARN'),
+    ).toBeInTheDocument();
```
JBWatenbergScality: I think this is likely not testing a specific use case. IIUC here you are checking that when mouse enter the 4th button of the page an element is present somewhere in the screen with the "Copy ARN" text, which is already tested just above. What do you wanted to test here ? A tooltip ? On which button ? Could we localise the button in a better way ? Also if it is a test concerning a tooltip we should localize it more precisely than with "Copy ARN" on the global screen as we saw that just above there is another component with the same text that would match.

src/react/account/__tests__/AccountUserList.test.tsx
```diff
@@ -124,5 +140,88 @@ describe('AccountUserList', () => {
...
+
+    const eyeButtonOfFirstRow = getAllByRole(firstRow, 'button')[0];
+    expect(
+      eyeButtonOfFirstRow.getElementsByClassName('fas fa-eye')[0].className,
```
JBWatenbergScality: ClassNames are not a reliable selector, please rely on accessible labels instead

src/react/account/__tests__/AccountUserList.test.tsx
```diff
@@ -124,5 +140,88 @@ describe('AccountUserList', () => {
...
+      /Copy ARN/i,
+    )).toBeInTheDocument();
+
+    fireEvent.mouseEnter(screen.getAllByRole('button')[4]);
+
+    expect(
+      screen.getByText('Copy ARN'),
+    ).toBeInTheDocument();
+
+    const eyeButtonOfFirstRow = getAllByRole(firstRow, 'button')[0];
```
JBWatenbergScality: We should better select button by their text as well or accessible labels when we can have several

src/react/account/__tests__/AccountUserList.test.tsx
```diff
@@ -124,5 +140,88 @@ describe('AccountUserList', () => {
...
+
+  });
+  it('should render header buttons and column names', async () => {
+
+    render(<AccountUserList accountName="account" />, {
+      wrapper,
+    });
+
+    fireEvent.click(screen.getAllByText('Loading users...')[0])
+    await waitFor(() => screen.getAllByRole('button')[2])
```
JBWatenbergScality: We should better select button by their text as well or accessible labels when we can have several

src/react/account/__tests__/AccountUserList.test.tsx
```diff
@@ -124,5 +140,88 @@ describe('AccountUserList', () => {
...
+    )).toBeInTheDocument();
+
+  });
+  it('should render header buttons and column names', async () => {
+
+    render(<AccountUserList accountName="account" />, {
+      wrapper,
+    });
+
+    fireEvent.click(screen.getAllByText('Loading users...')[0])
```
JBWatenbergScality: Why clicking on a loading users text ? What do you aim to test here ?

src/react/account/__tests__/AccountUserList.test.tsx
```diff
@@ -124,5 +140,88 @@ describe('AccountUserList', () => {
...
+
+    /**********           Number of columns :         ************/
+
+    expect(screen.getAllByRole('columnheader').length).toEqual(nbrOfColumnsExpectedWithoutEdit);
+
+    /**********           Buttons 'search' and 'Create user' exist :         ************/
+    const searchZone = screen.getAllByRole('button')[0];
+    expect(
+      searchZone.getElementsByClassName('fas fa-search')[0].className,
+    ).toBeDefined();
```
JBWatenbergScality: We should better select button by their text as well or accessible labels when we can have several.

Eg. `screen.getByRole('button', {label: 'search'})`

src/react/account/__tests__/AccountUserList.test.tsx
```diff
@@ -124,5 +140,88 @@ describe('AccountUserList', () => {
...
+
+    const createButton = screen.getAllByRole('button')[2];
+
+    expect(getByText(
+      createButton,
+      new RegExp('Create User'),
+    )).toBeInTheDocument();
```
JBWatenbergScality: Same as above We should better select button by their text as well or accessible labels when we can have several.

src/react/account/__tests__/AccountUserList.test.tsx
```diff
@@ -124,5 +140,88 @@ describe('AccountUserList', () => {
...
+      ));
+
+    render(<AccountUserList accountName="account" />, {
+      wrapper,
+    });
+
+    await waitFor(() => screen.getByText('We failed to retrieve users, please retry later. If the error persists, please contact your support.'));
+    expect(screen.getByText('We failed to retrieve users, please retry later. If the error persists, please contact your support.')).toBeInTheDocument();
+
+
```
JBWatenbergScality:
```suggestion
```

src/react/utils/test.tsx
```diff
@@ -63,6 +93,42 @@ export const reduxMount = (component, testState) => {
...
+  return {
+    component: render(
+      <Wrapper>
+        <ThemeProvider theme={theme}>
+          <Provider store={store}>{component}</Provider>
+        </ThemeProvider>
+      </Wrapper>
+    ),
+  };
+};
```
JBWatenbergScality: This should be part of locations testing pr

JBWatenbergScality (changes requested):

JBWatenbergScality (changes requested):
I think some tests are missing regarding access keys retrieval and edit of an account

AsmaaELM:
Testing account users

PR #284, opened 2022-03-21
Delete users with refactor query

AsmaaELM:
Issue:  https://scality.atlassian.net/browse/ZKUI-74

### DIFF
--- a/src/react/account/AccountUserList.tsx
+++ b/src/react/account/AccountUserList.tsx
@@ -1,4 +1,5 @@
-import { ChangeEvent, useMemo } from 'react';
+import React, { ChangeEvent, useMemo, useState } from 'react';
+import styled from 'styled-components';
 import { useHistory, useRouteMatch } from 'react-router-dom';
 import { Table, Button } from '@scality/core-ui/dist/next';
 import TextBadge from '@scality/core-ui/dist/components/textbadge/TextBadge.component';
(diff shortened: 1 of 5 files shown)


ChengYanJin:
please target development/1.5

src/react/account/AccountUserList.tsx
```diff
@@ -126,6 +138,49 @@ const WithTooltipWhileLoading = ({
   </>
 );
 
+const DeleteUserAction = (rowValue: { userName : string} , accountName: string) => {
+  const { userName } = rowValue;
+  const IAMClient = useIAMClient();
+  const [showModal, setShowModal] = useState(false);
+  const { data: accessKeysResult } = useAwsPaginatedEntities(getUserAccessKeysQuery(userName, notFalsyTypeGuard(IAMClient)), data => data.AccessKeyMetadata);
+  const { data: listGroupsResult } = useAwsPaginatedEntities(getUserListGroupsQuery(userName, notFalsyTypeGuard(IAMClient)), data => data.Groups);
```
JBWatenbergScality: We likely should disable the button while we haven't loaded those entities. Also we should handle error cases.
While the entities are loading and the button is disabled we should display a tooltip mentioning the loading/checking step. 
In addition perf could be a bit optimised because in theory we just need the first page to know if the Account is linked to some entities and don't need to load all the entities. So maybe a simple useQuery would have been enough here.

JBWatenbergScality (changes requested):

src/react/account/AccountUserList.tsx
```diff
@@ -126,6 +138,49 @@ const WithTooltipWhileLoading = ({
...
+
+  return (
+    <>
+      <DeleteConfirmation
+        show={showModal}
+        cancel={() => setShowModal(false)}
+        approve={() => {
+          deleteUserMutation.mutate(userName);
+        }}
+        titleText={`Delete User Key? \n Permanently remove the following user ${userName} ?`}
```
ChengYanJin: According to @Cuervino 's design (https://scality.atlassian.net/browse/ZKUI-32)
```
Permanently remove the following IAM User?
${userName}
```

ChengYanJin: And the new `Delete` button should be just next to the `Copy ARN` and `Edit` buttons.

src/react/account/AccountUserList.tsx
```diff
@@ -200,6 +256,16 @@ const AccountUserList = ({ accountName }: { accountName?: string }) => {
...
+    {
+      Header: '',
+      accessor: 'deleteAction',
+      cellStyle: {
+        textAlign: 'right',
+        minWidth: '5rem',
+      },
+      disableSortBy: true,
+      Cell: (value: CellProps) => DeleteUserAction(value.row.original, notFalsyTypeGuard(accountName)),
+    },
```
ChengYanJin: ideally we should have this button in the same column of actions. 
```js
const renderActionButtons = (rowValues) => {
  const { arn, userName } = rowValues;
  return (
    <div style={{ display: 'flex' }}>
      <CopyARNButton text={arn} />
      <RenderEditButton userName={userName} />
      <RenderDeleteButton />  // to be added
    </div>
  );
};
```

ChengYanJin (changes requested):

src/react/account/AccountUserList.tsx
AsmaaELM: Thank you for your comment. Is it mandatory to use useQuery in this case, because I tried to use it but I noticed that the delete button stayed enabled all the time?  If it should be used can you please tell me what's wrong with my query at the line 145 in AccountUserList.tsx ? Thanks!

JBWatenbergScality (approved):

AsmaaELM:
ping

ChengYanJin (approved):

PR #287, opened 2022-03-22
update select buckets in the new workflow

AsmaaELM:
Issue: https://scality.atlassian.net/browse/ZKUI-111
https://scality.atlassian.net/browse/ARTESCA-1558

### DIFF
--- a/src/react/workflow/ExpirationForm.tsx
+++ b/src/react/workflow/ExpirationForm.tsx
@@ -2,7 +2,8 @@ import * as T from '../ui-elements/TableKeyValue2';
 import { Controller, useFormContext } from 'react-hook-form';
 import { ErrorInput } from '../ui-elements/FormLayout';
 import type { Locations } from '../../types/config';
-import { Select, Toggle, SpacedBox } from '@scality/core-ui';
+import { Toggle, SpacedBox } from '@scality/core-ui';
+import Select, { Option } from '@scality/core-ui/dist/components/selectv2/Selectv2.component';
 import { flattenFormErrors, renderSource, sourceBucketOptions } from './utils';
 import Joi from '@hapi/joi';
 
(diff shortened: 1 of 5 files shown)


JBWatenbergScality (changes requested):
It should use the selectv2 from core-ui

AsmaaELM:
ping

src/react/workflow/ReplicationForm.tsx
```diff
@@ -196,22 +187,22 @@ function ReplicationComponent({
...
-                        options={options}
-                        formatOptionLabel={renderSource(locations)}
-                        isDisabled={isEditing}
-                        isOptionDisabled={(option) => option.disabled === true}
-                        value={result}
-                      />
+                      <Select value={result} onChange={onChange} isDisabled={isEditing}>
+                        {options &&
+                          options.map((o, i) => (
+                            <Option title={o.disabled ? 'Versioning is disabled' : ''} key={i} value={o.value}
```
ChengYanJin, line 193: I am wondering do we need the `title` for each of the options?

AsmaaELM: Well, we need the title with the value  "Versioning is disabled" just in case of a disabled option

src/react/workflow/ExpirationForm.tsx
```diff
@@ -164,14 +165,15 @@ export function ExpirationForm({ bucketList, locations, prefix = '' }: Props) {
...
-                        onChange={({ value }) => onChange(value)}
-                        options={options}
-                        formatOptionLabel={renderSource(locations)}
-                        isOptionDisabled={(option) => option.disabled === true}
-                        value={result}
-                      />
+                      <Select id='sourceBucket' value={result} onChange={onChange}>
+                        {options &&
+                          options.map((o, i) => (
+                            <Option title={o.disabled ? 'Versioning is disabled' : ''} key={i} value={o.value}
```
JBWatenbergScality: For expiration workflows this is not necessary
```suggestion
                            <Option key={i} value={o.value}
```

AsmaaELM: Ok, I remove it

PR #288, opened 2022-03-22
Improvement/zk UI 116 react query delete wkf

JBWatenbergScality:


package-lock.json
```diff
@@ -68,6 +68,7 @@
         "@testing-library/jest-dom": "^5.16.2",
         "@testing-library/react": "^12.1.3",
         "@testing-library/react-hooks": "^7.0.2",
+        "@types/hapi__joi": "^17.1.8",
```
AsmaaELM, line 71: Did you push the package-lock.json  file on purpose?

src/react/workflow/ConfigurationTab.tsx
```diff
@@ -210,8 +243,18 @@ function ConfigurationTab({
       <FormProvider {...useFormMethods}>
         <FormContainer style={{ backgroundColor: 'transparent' }}>
           <form onSubmit={handleSubmit(onSubmit)}>
+            <T.Group>
```
AsmaaELM, line 246: Is it possible in this case to use the new version of the Table?

src/react/workflow/ConfigurationTab.tsx
```diff
@@ -223,7 +266,7 @@ function ConfigurationTab({
                   marginRight: spacing.sp24,
                 }}
                 variant="outline"
-                onClick={() => history.push('./')}
+                onClick={() => history.replace()}
```
AsmaaELM, line 269: I think that you should provide a path to replace. Or maybe you can try history.goBack() instead ?

src/react/workflow/ReplicationForm.tsx
```diff
@@ -37,14 +44,32 @@ const ReplicationContainer = styled.div`
...
   bucketList: S3BucketList;
   locations: Locations;
-  workflow: ReplicationStream | null | undefined;
+  workflow?: ReplicationStream;
 };
 
+export const replicationSchema = Joi.object({
+  streamId: Joi.string().label('Id').allow(''),
+  streamVersion: Joi.number().label('Version').optional(),
+  // streamName: Joi.string().label('Name').min(4).allow('').messages({
```
AsmaaELM, line 56: Is it necessary to keep this comment or can we delete it?

src/react/workflow/ConfigurationTab.tsx
```diff
@@ -64,65 +58,101 @@ const ConfigurationHeader = styled.div`
...
+  const [isDeleteModalOpen, setIsDeleteModalOpen] = useState(false);
+  const mgnt = useManagementClient();
+  const state = useSelector((state: AppState) => state);
+  const { instanceId } = getClients(state);
+  const accountId = getAccountId(state);
+
+  const replicationsQuery = useQuery({
+    ...workflowListQuery(
+      notFalsyTypeGuard(mgnt),
+      accountId,
```
AsmaaELM, line 77: I think we should use notFalsyTypeGuard for both accountId and instanceId because they can be null

src/react/workflow/ReplicationForm.tsx
```diff
@@ -54,6 +79,22 @@ function ReplicationComponent({
 
   const { register, errors, control, reset, getValues } = useFormContext();
 
+  const state = useSelector((state: AppState) => state);
```
AsmaaELM, line 92: I think the import of AppState is missing in this class

src/react/actions/__tests__/replication.test.ts
```diff
@@ -1,28 +0,0 @@
-import * as actions from '../replication';
```
AsmaaELM, line 1: There are some failed tests like in workflow.test.ts, BucketCreate.test.tsx, ObjectLockSetting.test.tsx ...
Can you please take a look at them ?

src/react/workflow/ConfigurationTab.tsx
```diff
@@ -64,65 +58,101 @@ const ConfigurationHeader = styled.div`
...
 }: Props) {
   const dispatch = useDispatch();
   const history = useHistory();
   const queryClient = useQueryClient();
   const { workflowId } = wfSelected;
-  const isDeleteModalOpen = useSelector(
-    (state: AppState) => state.uiWorkflows.showWorkflowDeleteModal,
-  );
+  const [isDeleteModalOpen, setIsDeleteModalOpen] = useState(false);
+  const mgnt = useManagementClient();
```
AsmaaELM, line 69: Can we use a more significant name or do you think it's ok to use "mgnt"?

src/react/workflow/ConfigurationTab.tsx
```diff
@@ -64,65 +58,101 @@ const ConfigurationHeader = styled.div`
...
+  const state = useSelector((state: AppState) => state);
+  const { instanceId } = getClients(state);
+  const accountId = getAccountId(state);
+
+  const replicationsQuery = useQuery({
```
AsmaaELM, line 74: Can we move this query to queries.ts in order to refactor and avoid duplication in ConfigurationTab.tsx and ReplicationForm.tsx?

package-lock.json
JBWatenbergScality: Yes pushing the package-lock.json file is important to fix dependencies used by others and by the CI (https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json)

src/react/workflow/ConfigurationTab.tsx
JBWatenbergScality: No, this is a form build using a table which is not the use case addressed by the "data" table of core-ui

src/react/workflow/ConfigurationTab.tsx
JBWatenbergScality: This is not what I intend to do, here I don't want to redirect the user to the previous page, I want a "soft page reload" to force cleanup of the form. For now this approach was necessary but following to react hook form we may have a cleaner way to handle this, I will check this when implementing lifecycle forms

src/react/workflow/ReplicationForm.tsx
JBWatenbergScality: Yes we are going to temporary keep it as we are studying why it has been commented in the first place and may weant to very soon re-offer the ability to set a name for the workflows

src/react/actions/__tests__/replication.test.ts
JBWatenbergScality: Done in https://github.com/scality/zenko-ui/pull/288/commits/0608edde334f62e9c547044730cdd0647e96d9b4 it was due to the bump of react-hook-form and the old test too tightly coupled with the implementation, I ended up rewriting most of those tests.

src/react/workflow/ConfigurationTab.tsx
JBWatenbergScality: I think it is fine for now, I'm going to improve and cleanup a bit the code when implementing lifecycle

src/react/workflow/ConfigurationTab.tsx
JBWatenbergScality: The query is already mutualised we are using various selectors in 3 different places ConfigurationTab, ReplicationForm and Workflows. It may make sense to factorise this selector through `(workflows) => workflows.filter(w => w.replication).map(w => w.replication)` I will check that with the next lifecycle implem

src/react/workflow/ConfigurationTab.tsx
JBWatenbergScality: I prefer not taking this approach and refactor the way we retrieved these info so that their typing is actually the expected one at the level of this component, however as this may take a bit of time that I don't have for now, I think it is fine to merge as it is now

src/react/workflow/ReplicationForm.tsx
JBWatenbergScality: good catch I just added it

AsmaaELM (approved):

PR #289, opened 2022-03-23
improve new workflow table

AsmaaELM:
Issue: https://scality.atlassian.net/browse/ZKUI-117

### DIFF
--- a/src/react/workflow/WorkflowList.tsx
+++ b/src/react/workflow/WorkflowList.tsx
@@ -1,147 +1,119 @@
-import MemoRow, { createItemData } from './WorkflowRow';
-import Table, * as T from '../ui-elements/Table';
-import { useFilters, useFlexLayout, useSortBy, useTable } from 'react-table';
-import { AutoSizer } from 'react-virtualized';
-import { FixedSizeList } from 'react-window';
-import { ListSection } from '../ui-elements/ListLayout3';
+import React, { useCallback } from 'react';
 import { TextTransformer } from '../ui-elements/Utility';
 import type { Workflows } from '../../types/workflow';
-import { useDispatch, useSelector } from 'react-redux';
 import { useHistory } from 'react-router-dom';
-import { AppState } from '../../types/state';
-const columns = [
-  {
-    Header: 'Workflow Description',
-    accessor: 'name',
-    width: 60,
-  },
-  {
-    Header: 'Workflow Type',
-    accessor: 'type',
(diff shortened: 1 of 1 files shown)


AsmaaELM:
ping

src/react/workflow/WorkflowList.tsx
```diff
@@ -1,147 +1,131 @@
...
+      <span> {`${row.values.name} ${data}`} </span>
+    );
+  }
+
+  function RowAsync({ row } : {row: RowType}) {
+    const [loading, setLoading] = useState(true);
+    const [data, setData] = useState('');
+
+    useEffect(() => {
+        setData('loading data');
```
ChengYanJin: I think we don't need to async load the name since it already exists in the `workflows` which we pass to component WorkflowList

AsmaaELM: Done!

src/react/workflow/WorkflowList.tsx
```diff
@@ -1,147 +1,128 @@
...
+    useEffect(() => {
+      setLoading(false);
+    }, [loading]);
+
+    return (
+      <DataComponent
+        row={row}
+        loading={loading}
+      />
+    );
```
JBWatenbergScality:
```suggestion
    return (
      <DataComponent
        row={row}
      />
    );
```

src/react/workflow/WorkflowList.tsx
```diff
@@ -1,147 +1,128 @@
...
+  }
+  const renderRowSubComponent = useCallback(
+    ({ row }) => <RowAsync row={row} />,
+[]);
```
JBWatenbergScality: Can you format this code using prettier?

src/react/workflow/WorkflowList.tsx
AsmaaELM: Done!

src/react/workflow/WorkflowList.tsx
AsmaaELM: Done!

JBWatenbergScality (changes requested):
image
The scrollable area should expand to the whole area

AsmaaELM:
Screenshot 2022-03-30 at 10 14 59

AsmaaELM:
CSS modified

JBWatenbergScality (approved):

PR #290, opened 2022-03-25
Expiration lifecycle support

JBWatenbergScality:
To be fixed:
We skipped the test for Locations, because the API call is not correctly being intercepted by MSW


src/react/workflow/Workflows.tsx
```diff
@@ -15,6 +15,36 @@ import { notFalsyTypeGuard } from '../../types/typeGuards';
...
+      expirations: workflows
+        .filter((w) => w.expiration)
+        .map((w) => w.expiration),
+    }),
+  });
+
+  return workflowsQuery;
+}
 
 export default function Workflows2() {
```
ChengYanJin: I think we can rename it to just `Workflow`?

src/react/workflow/utils.tsx
```diff
@@ -151,10 +169,36 @@ export function convertToReplicationStream(
     },
   };
 }
+export function prepareExpirationQuery(data: Expiration): Expiration {
```
ChengYanJin, line 172: Would be great to add some tests to cover the 2 utils functions! ;)

ChengYanJin (approved):

PR #294, opened 2022-03-30
css polishing

AsmaaELM:
Issue: https://scality.atlassian.net/browse/ZKUI-120

### DIFF
--- a/src/react/ui-elements/TableKeyValue2.tsx
+++ b/src/react/ui-elements/TableKeyValue2.tsx
@@ -57,6 +57,10 @@ export const Groups = styled.div`
 
   max-width: 450px;
 `;
+export const ScrollArea = styled.div`
+  overflow: auto;
+  max-height: calc(100vh - 19rem);
+`;
 export const Group = styled.div`
   margin-bottom: ${spacing.sp24};
 `;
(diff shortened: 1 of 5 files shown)


src/react/ui-elements/TableKeyValue2.tsx
```diff
@@ -57,6 +57,10 @@ export const Groups = styled.div`
 
   max-width: 450px;
 `;
+export const ScrollArea = styled.div`
+  overflow: auto;
+  max-height: calc(100vh - 19rem);
```
ChengYanJin, line 62: On my side, the Cancel and Save button are still not visible by default...

ChengYanJin (changes requested):

src/react/ui-elements/TableKeyValue2.tsx
AsmaaELM: I just tested.  on my branch I get this result:
Screenshot 2022-03-30 at 19 15 03
And in development, we have this result : 
Screenshot 2022-03-30 at 19 15 48

Or  maybe Yanjin, you're talking about another screen  🤔 ?

AsmaaELM: Here is the expiration edition after css modification:
Screenshot 2022-03-31 at 10 10 06

ChengYanJin: Hi @AsmaaELM 
Do you think we can avoid the empty space on the right side of the screen? 
by increasing the width of the workflow table maybe?

ChengYanJin (approved):

PR #295, opened 2022-03-30
ZKUI-118: compute workflow rule name in the ui

JBWatenbergScality:


Cuervino (approved):
I cannot find any comment to make, looks like a good naming policy for me :-)

PR #296, opened 2022-03-30
ZKUI-121: Add workflow tab in the bucket list page

ChengYanJin:
image


src/react/databrowser/buckets/details/Workflow.tsx
```diff
@@ -0,0 +1,120 @@
...
+  const select = (workflows: APIWorkflows) => makeWorkflows(workflows);
+  const { data, status } = useWorkflows(select, [bucketName]);
+
+  const nameCell = (value) => {
+    const id = value.row.original.id;
+    const workflowName = value.value;
+    return (
+      <NameLinkContaner
+        onClick={() =>
+          dispatch(push(`/accounts/${accountName}/workflows/${id}`))
```
JBWatenbergScality: we don't need to dispatch here. We can simply do 
```suggestion
          history.push(`/accounts/${accountName}/workflows/${id}`)
```

JBWatenbergScality (approved):

PR #300, opened 2022-03-31
select account on create workflow

AsmaaELM:
Issue:  https://scality.atlassian.net/browse/ZKUI-125

### DIFF
--- a/src/react/ui-elements/Breadcrumb.tsx
+++ b/src/react/ui-elements/Breadcrumb.tsx
@@ -10,14 +10,14 @@ import styled from 'styled-components';
 import { useDispatch } from 'react-redux';
 import { useQueryParams } from '../utils/hooks';
 import { ellipsis } from 'polished';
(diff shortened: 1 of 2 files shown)


AsmaaELM:
ping

ChengYanJin (approved):
LGTM!
Not related to your PR, but I found an issue with the account selector, when I scroll down it's getting blurring.. we should investigate this issue.

ChengYanJin:
there is little shifting because the account selector in /workflow and in /create-workflow, it might be a quick fix ;)

AsmaaELM:
> there is little shifting because the account selector in /workflow and in /create-workflow, it might be a quick fix ;)

Yeah, thanks for the info. I fix it by increasing the width in the Breadcrumb.tsx

AsmaaELM:
ping

AsmaaELM:
ping

PR #302, opened 2022-03-31
ZKUI-126: Filter out the scality internal account

ChengYanJin:
Create a hook `useAccount()` to retrieve the accounts from configurations and filter out the account `scality-internal-account`.

### DIFF
--- a/src/react/account/AccountContent.tsx
+++ b/src/react/account/AccountContent.tsx
@@ -6,27 +6,25 @@ import {
   useParams,
   useRouteMatch,
 } from 'react-router-dom';
-import { useSelector } from 'react-redux';
 import * as L from '../ui-elements/ListLayout5';
 import AccountDetails from './AccountDetails';
 import AccountHead from './AccountHead';
-import type { AppState } from '../../types/state';
 import { BreadcrumbAccount } from '../ui-elements/Breadcrumb';
 import AccountCreateUser from './AccountCreateUser';
 import AccountUpdateUser from './AccountUpdateUser';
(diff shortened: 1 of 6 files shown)


src/react/account/AccountContent.tsx
```diff
@@ -17,16 +17,16 @@ import AccountUpdateUser from './AccountUpdateUser';
 import AccountUserAccessKeys from './AccountUserAccessKeys';
 import CreateWorkflow from '../workflow/CreateWorkflow';
 import Workflows from '../workflow/Workflows';
+import { useAccounts } from '../utils/hooks';
```
AsmaaELM, line 18: I think we can remove both useSelector and useAccounts imports because we're not using them anymore

src/react/utils/hooks.ts
```diff
@@ -84,4 +86,14 @@ export const useClipboard = () => {
...
-};
\ No newline at end of file
+};
+
+export const useAccounts = () => {
+  const accounts = useSelector(
+    (state: AppState) => state.configuration.latest.users,
+  );
+
+  return accounts.filter(
```
AsmaaELM, line 96: Maybe we can filter by account Id because the userName can be edited one day ?

ChengYanJin: Hi @AsmaaELM 
Just in case, I filter out the account by either the userName or the id.

AsmaaELM (approved):
Approved!

JBWatenbergScality (approved):

ChengYanJin:
ping

PR #329, opened 2022-04-22
ARTESCA-3994: Support metadatasearch when ListVersioning is on

ChengYanJin:


ChengYanJin:
wrong branch name

src/react/actions/s3object.ts
```diff
@@ -198,11 +198,11 @@ export function createFolder(
   };
 }
 
-function _getObjectListWithSignedUrlAndLockStatus<
+export function _getObjectListWithSignedUrlAndLockStatus<
```
JBWatenbergScality, line 201: Nit: if we make it a public function we should remove the `_` prefixing the name
```suggestion
export function getObjectListWithSignedUrlAndLockStatus<
```

src/react/actions/zenko.ts
```diff
@@ -128,12 +135,81 @@ function _getSearchObjects(
...
+          DeleteMarker,
+          CommonPrefixes,
+          Prefix,
+        }: SearchBucketVersionsResp) => {
+          const list =
+            await _getObjectListWithSignedUrlAndLockStatus<S3Version>(
+              zenkoClient,
+              bucketName,
+              Version,
+            );
```
JBWatenbergScality, line 170: I'm not sure to understand why we are listing objects of the bucket here ?

src/react/actions/zenko.ts
```diff
@@ -128,12 +135,81 @@ function _getSearchObjects(
...
+          if (NextKeyMarker) {
+            return dispatch(
+              continueListObjectVersionsSuccess(
```
JBWatenbergScality, line 173: Will this really re-trigger a search if the search contains a lot of items ? I'm afraid not, I think It will try to fetch a second page of objects but using the marker of the search. Likely a bug here

src/react/databrowser/objects/ObjectList.tsx
```diff
@@ -45,13 +45,16 @@ export default function ObjectList({
...
+
   const errorZenkoMsg = useSelector(
     (state: AppState) => state.zenko.error.message,
   );
   const isBucketVersioningDisabled = isVersioningDisabled(
     bucketInfo.versioning,
   );
-  const isMetadataType = listType === LIST_OBJECTS_METADATA_TYPE;
+
+  const isMetadataType = listType === LIST_OBJECTS_METADATA_TYPE || searchInput;
```
JBWatenbergScality, line 57: Nit: It likely could be reduced to
```suggestion
  const isMetadataType = !!searchInput;
```
And we could drop the LIST_OBJECTS_METADATA_TYPE logic

JBWatenbergScality: This can be done in a second iteration

src/types/zenko.ts
```diff
@@ -27,6 +38,15 @@ export type SearchBucketResp = {
...
 };
+export type SearchBucketVersionsResp = {
+  readonly IsTruncated?: boolean;
+  readonly NextKeyMarker?: Marker;
+  readonly NextVersionIdMarker?: Marker;
+  readonly Version: Array<S3Version>;
+  readonly DeleteMarker: Array<S3DeleteMarker>;
+  readonly commonPrefixes: CommonPrefix;
+  readonly Prefix: string;
+};
```
JBWatenbergScality, line 49: Nit: I think the searchBucketVersionsResp is Exactly the same type as the listObjectVersion so we might want to use AWS sdk ListObjectVersionResponse type here.

src/types/zenko.ts
```diff
@@ -1,13 +1,24 @@
...
   readonly Bucket: string;
   readonly Query: string;
   readonly Marker?: string;
 };
+export type SearchVersionParams = {
+  readonly Bucket: string;
+  readonly Query: string;
+  readonly KeyMarker?: string;
+  readonly VersionIdMarker?: string;
+};
```
JBWatenbergScality, line 21: Could be 

```suggestion
type CommonSearchParams = {
  readonly Bucket: string;
  readonly Query: string;
}
export type SearchParams = CommonSearchParams & {
  readonly Marker?: string;
};
export type SearchVersionParams = CommonSearchParams & {
  readonly KeyMarker?: string;
  readonly VersionIdMarker?: string;
};
```

src/react/actions/zenko.ts
ChengYanJin: Indeed it's not necessary here.

src/types/zenko.ts
ChengYanJin: I do agree with we could benefit from the type `ListObjectVersionOutput` in `aws-sdk/clients/s3`.
However, I found the output  structure in zenkoclient I have to define Version/DeleteMarker instead of the type inn the AWS SDK Version**s**/DeleteMarker**s**

src/react/actions/zenko.ts
ChengYanJin: Indeed, I think in the end we only need `listObjectVersionsSuccess` action here.. 
and the continueListObject login will be handled within the table

src/types/zenko.ts
JBWatenbergScality: Hmm that looks like a bug to me, I think the whole idea of that endpoint was to be compatible with the listVersionCall. For example an application using AWS SDK should be able to listObjects and pass the additional search params without changing the response parsing code. @nicolas2bert can you confirm this ? Is this response difference intentional ?

PR #332, opened 2022-04-25
Feature/zk UI 141 handle errors and loading states

JBWatenbergScality:


src/react/account/__tests__/Accounts.test.tsx
```diff
@@ -0,0 +1,113 @@
+import {
+  screen,
+  waitFor,
+  waitForElementToBeRemoved,
```
ChengYanJin: can be removed ;)

ChengYanJin (approved):
LGTM! It's cool to initial this way to handle errors!

src/react/account/__tests__/Accounts.test.tsx
JBWatenbergScality: done

PR #334, opened 2022-04-26
ZKUI-143: Assume the right role when selecting an account

ChengYanJin:
- move the data browser route under the /accounts
- get rid of selectAccountID action and create context to access the current selected account
- temporarily remove the root user email ad root user ARN from the account properties tab since we stop calling pensive-api


src/js/IAMClient.ts
```diff
@@ -1,30 +1,27 @@
...
   const { stsClient } = getClients(state);
-  const accounts = configuration.latest.users;
-  const account = accounts.find((a) => a.Name === accountName);
-  if (!account || !oidc || !oidc.user) return Promise.reject();
-  const assumeRoleParams = getAssumeRoleWithWebIdentityParams(oidc, account.id);
+
+  const assumeRoleParams = {
+    idToken: oidc.user.id_token,
+    roleArn: roleArn,
+    RoleSessionName: oidc.user.profile.sub,
```
ghost: I would suggest using an identifier in the session name, so that we know it's coming from the UI. What do you think?

src/js/IAMClient.ts
```diff
@@ -1,30 +1,27 @@
...
 import type { Credentials } from '../types/zenko';
 import IAM from 'aws-sdk/clients/iam';
-import type { IAMClient as IAMClientInterface, WebIdentityRoles } from '../types/iam';
+import type {
+  IAMClient as IAMClientInterface,
+  WebIdentityRoles,
+} from '../types/iam';
 import { getClients } from '../react/utils/actions';
 import { notFalsyTypeGuard } from '../types/typeGuards';
 export const rolePathName = 'scality-internal/storage-manager-role';
```
JBWatenbergScality: Is this "default role" still in use by anything ? It should probably not be the case

src/react/utils/hooks.ts
```diff
@@ -186,6 +187,8 @@ export const useAccounts = () => {
           Name: current.Name,
           CreationDate: current.CreationDate,
           Roles: [...(agg[current.Name]?.Roles || []), ...current.Roles],
+          // retrieve the account id from the role ARN
+          id: current.Roles[0].Arn.split(':')[4],
```
JBWatenbergScality: nit: I would suggest parsing the ARN with a regex to get clearer and more resiliently the Id.

`arn:aws:iam::000000000000:role/scality-internal/storage-manager-role` could be parsed using

`/arn:aws:iam::(?<account_id>/d+):/(?<role_name>.+)/` then you can retrieve the id using the `account_id` capture group.

src/react/account/details/properties/AccountInfo.tsx
```diff
@@ -91,7 +91,8 @@ function AccountInfo({ account }: Props) {
             <T.Key> Creation Date </T.Key>
             <T.Value> {formatDate(new Date(account.createDate))} </T.Value>
```
JBWatenbergScality: Might be CreationDate field when we get this info from `getRolesFromWebIdentity`

src/react/utils/hooks.ts
```diff
@@ -199,3 +200,14 @@ export const useAccounts = () => {
...
+
+export const useInvalidAccountsQuery = () => {
+  const users = useSelector(
+    (state: AppState) => state.configuration.latest.users,
+  );
+  const queryClient = useQueryClient();
+  const token = useSelector((state: AppState) => state.oidc.user?.access_token);
+  useEffect(() => {
+    queryClient.invalidateQueries(['WebIdentityRoles', token]);
+  }, [users.length]);
```
JBWatenbergScality: I'm not very sure to understand the intent of this code. We should invalidate WebIdentityRoles when we delete an account or create a new one, but for this we should probably not based the invalidation on the configuration overlay as we are studying it's removal

src/react/account/details/properties/__tests__/AccountInfo.test.tsx
```diff
@@ -48,17 +49,17 @@ describe('AccountInfo', () => {
...
+    //   key: 'Root User Email',
+    //   value: account1.email,
+    //   extraCellComponent: 'Clipboard',
+    // });
+    // const sixthRow = rows.at(4);
+    // testRow(sixthRow, {
+    //   key: 'Root User ARN',
+    //   value: account1.arn,
+    //   extraCellComponent: 'Clipboard',
+    // });
```
JBWatenbergScality, line 63: nit: we likely would want to mutualize it in the same commit as the one that comment out the Root user email and arn but I won't bother too much about that ;)

src/react/account/__tests__/AccountDetails.test.tsx
```diff
@@ -48,6 +48,7 @@ describe('AccountDetails', () => {
             pathname: '/accounts/bart',
           },
         },
+        oidc: { user: { id_token: 'id_token', profile: { sub: 'sub' } } },
```
JBWatenbergScality: We probably could add this in the global init state for all the tests.

src/react/account/details/properties/AccountInfo.tsx
JBWatenbergScality: Ok done in fe638b016fc31883f50b55a6858c22572029a889

src/react/account/details/properties/__tests__/AccountInfo.test.tsx
ChengYanJin: Done! All in one commit https://github.com/scality/zenko-ui/pull/334/commits/464d4093da4966a6961e9a01327eff0c0a11a169

src/js/IAMClient.ts
ChengYanJin: Hi @williamlardier, 
Thanks again for the review!!
And sure! Fixed in https://github.com/scality/zenko-ui/pull/334/commits/ab956ce2a70fa2f33b5859c440bc9758c1f1b89b

src/js/IAMClient.ts
ChengYanJin: Indeed, getting rid of of the default hard-coded rolePathName in https://github.com/scality/zenko-ui/pull/334/commits/52f3f12f9ab44ab7a39d16f97b9ba045c383eacf

src/react/utils/hooks.ts
ChengYanJin: That's right we should probably invalidate `WebIdentityRoles` cache based on the local storage of the accountID.
https://github.com/scality/zenko-ui/pull/334/commits/0129c206bfe2b2db6840033ad8030efeebf2fb5f

src/react/account/__tests__/AccountDetails.test.tsx
ChengYanJin: Added here: https://github.com/scality/zenko-ui/pull/334/commits/bc737d8337f1d4de40258df1ab6ca2a113d7f8a3#diff-23d60af5da5b0ed4bd31d68462362b98adab06632006e787f9e74bc7eaf34790R173

src/react/utils/hooks.ts
ChengYanJin: Yes...for sure we should rely on regex!
Fixed here: 
https://github.com/scality/zenko-ui/pull/334/commits/0129c206bfe2b2db6840033ad8030efeebf2fb5f#diff-34f23db3cb2d457a06d84e4648f2ffe766e7f9b16d44c389f4e766bf2fa514a1R147

src/react/utils/hooks.ts
JBWatenbergScality: Sorry I still feel it is a bit overcomplicated/not really optimised. To me having the cache invalidation next to the action (creation or deletion) would be preferable

src/react/account/__tests__/AccountDetails.test.tsx
JBWatenbergScality: https://github.com/scality/zenko-ui/pull/334#commitcomment-72666210

JBWatenbergScality (approved):

PR #340, opened 2022-05-04
ZKUI-151: Redirect the user to data browser when assuming any other role than Storage Manager or Storage Account Owner

JBWatenbergScality:


ChengYanJin (approved):
LGTM!

PR #341, opened 2022-05-05
ZKUI-152: Handle token expired error

JBWatenbergScality:


ChengYanJin (approved):
LGTM! Let's go since this is the only way to handle this token expired error.

PR #342, opened 2022-05-05
ZKUI-147: Select Account and Role from the modal

ChengYanJin:
image

### DIFF
--- a/src/react/account/AccountRoleSelectButtonAndModal.tsx
+++ b/src/react/account/AccountRoleSelectButtonAndModal.tsx
@@ -1,11 +1,9 @@
 import { useState } from 'react';
-import { useDispatch, useSelector } from 'react-redux';
 import { useRouteMatch } from 'react-router';
 import Table from '@scality/core-ui/dist/components/tablev2/Tablev2.component';
 import { Button } from '@scality/core-ui/dist/next';
 import { Tooltip } from '@scality/core-ui';
-import { AppState } from '../../types/state';
-import { closeAccountRoleSelectModal } from '../actions';
+import SpacedBox from '@scality/core-ui/dist/components/spacedbox/SpacedBox';
 import { CustomModal as Modal, ModalBody } from '../ui-elements/Modal';
 import { regexArn, SCALITY_INTERNAL_ROLES, useAccounts } from '../utils/hooks';
 import {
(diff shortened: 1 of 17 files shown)


src/react/account/AccountRoleSelectModal.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+  const accountsWithRoles = [];
+  accounts.forEach((account) => {
+    const accountName = account.Name;
+    account.Roles.forEach((role) => {
+      const rolePathName = regexArn.exec(role.Arn)?.groups['role_name'] || '';
+      const rolePath = rolePathName.includes('/')
+        ? rolePathName.split('/')[0]
+        : '/';
+      const roleName = rolePathName.split('/')[1] || rolePathName;
```
JBWatenbergScality: I think a role path can include several sub paths. The role name would only be the last part of the arn (after the last /)

JBWatenbergScality: This is confirmed by the example here https://docs.aws.amazon.com/IAM/latest/APIReference/API_CreateRole.html

JBWatenbergScality: The regexArn could extract the path for you.

```
export const regexArn =
  /arn:aws:iam::(?<account_id>\d{12}):role\/(?<role_path>(?:[^\/]*\/)*)(?<role_name>[^\/]+)$/;
```

src/react/account/AccountRoleSelectModal.tsx
```diff
@@ -0,0 +1,167 @@
...
+    });
+  });
+
+  if (!isModalOpen) {
+    return null;
+  }
+
+  const handleClose = () => {
+    dispatch(closeAccountRoleSelectModal());
+  };
```
JBWatenbergScality: I think we could simplify a bit this and remove the redux logic by collocating the modal with it's opening button (I mean within the same component). Then we wouldn't need redux, just render the ButtonWithModal component and we get the whole logic for free

JBWatenbergScality: This ButtonWithModal component could simply be the `AccountSelectorButton`

JBWatenbergScality (changes requested):

src/react/account/AccountRoleSelectButtonAndModal.tsx
```diff
@@ -0,0 +1,171 @@
...
+  const { roleArn } = useDataServiceRole();
+  const [assumedRoleArn, setAssumedRoleArn] = useState(roleArn);
+  const [currentAccountName, setCurrentAccountName] = useState(accountName);
+  const [isModalOpen, setIsModalOpen] = useState(false);
+  const accountsWithRoles = [];
+  accounts.forEach((account) => {
+    const accountName = account.Name;
+    account.Roles.forEach((role) => {
+      const rolePath = regexArn.exec(role.Arn)?.groups['role_path'] || '';
+      const roleName = regexArn.exec(role.Arn)?.groups['role_name'] || '';
```
JBWatenbergScality: nit: regex could be executed only once as it can be a bit costly to execute
```
const parsedArn = regexArn.exec(role.Arn); 
const rolePath = parsedArn?.groups['role_path'] || '';
const roleName = parsedArn?.groups['role_name'] || '';
```

src/react/utils/hooks.ts
```diff
@@ -217,3 +228,11 @@ export const useAccounts = () => {
     (account) => account.Name !== 'scality-internal-services',
   );
 };
+
+export const useRolePathName = () => {
+  const { roleArn } = useDataServiceRole();
+  const rolePath = regexArn.exec(roleArn)?.groups['role_path'] || '';
+  const roleName = regexArn.exec(roleArn)?.groups['role_name'] || '';
```
JBWatenbergScality: nit: same as above

JBWatenbergScality (approved):

src/react/account/AccountRoleSelectButtonAndModal.tsx
ChengYanJin: Good point! Thanks for the comments!!
Addressed and fixed up in 5c16d6b9663802092f3f7864e1ee7c1f8ac842fc

PR #349, opened 2022-05-16
ZKUI-158: Update Pensieve API client

JBWatenbergScality:
Regenerate Pensieve API client with new Transition Workflow definition.

Steps:
 - Get the swagger spec of Pensieve API (PENSIEVE-API/swagger.json)
 - Load swagger specification on editor.swagger.io
 - Click Generate Client > typescript-fetch
 - Replace the files in src/js/managementClient by the generated ones
 - Replace the following code block:

```ts
export class RequiredError extends Error {
    name: "RequiredError"
    constructor(public field: string, msg?: string) {
        super(msg);
    }
}
```

By

```ts
export class RequiredError extends window.Error {
    name: "RequiredError"
    constructor(public field: string, msg?: string) {
        super(msg);
    }
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,5 +112,8 @@
     "vega-lite": "^4.0.0-beta.10",
     "vega-tooltip": "^0.28.0",
     "zenkoclient": "github:scality/zenkoclient#1.2.0"
+  },
+  "msw": {
+    "workerDirectory": "public/assets"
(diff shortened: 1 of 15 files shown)


ChengYanJin (approved):
LGTM!

PR #354, opened 2022-05-24
Feature/zk UI 165 creation dmf location

JBWatenbergScality:


src/react/backend/location/LocationDetails/LocationDetailsTapeDMF.tsx
```diff
@@ -0,0 +1,186 @@
...
+      </Fieldset>
+      <Fieldset>
+        <Label required htmlFor="endpoint">
+          Endpoint
+        </Label>
+        <Input
+          name="endpoint"
+          id="endpoint"
+          type="text"
+          placeholder="example: ws://path.to.my.dmf"
```
ChengYanJin, line 132: I guess we could drop `example:` since it's already a placeholder?

ChengYanJin: OK. As discussed with @Cuervino , it's suggested to have the example: presented.

ChengYanJin (approved):
LGTM!

PR #369, opened 2022-06-07
should not redirect the user to buckets when storage manager and no roles can be assumed

JBWatenbergScality:


ChengYanJin (approved):
LGTM!

PR #389, opened 2022-06-29
Fix create bucket

AsmaaELM:
Issue:  https://scality.atlassian.net/browse/ZKUI-202

### DIFF
--- a/src/react/databrowser/buckets/BucketCreate.tsx
+++ b/src/react/databrowser/buckets/BucketCreate.tsx
@@ -173,9 +173,8 @@ function BucketCreate() {
               <F.Input
                 type="text"
                 id="name"
-                {...register('name')}
+                {...register('name', { onChange: clearServerError })}
                 autoFocus={true}
-                onChange={clearServerError}
                 aria-invalid={!!errors.name}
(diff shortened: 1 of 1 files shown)


src/react/databrowser/buckets/BucketCreate.tsx
```diff
@@ -182,7 +182,6 @@ function BucketCreate() {
                 id="name"
                 {...register('name')}
                 autoFocus={true}
-                onChange={clearServerError}
```
JBWatenbergScality, line 178: Removing the clearServerError call on change is not the right way to solve the issue.

src/react/databrowser/buckets/BucketCreate.tsx
```diff
@@ -182,7 +182,6 @@ function BucketCreate() {
                 id="name"
                 {...register('name')}
```
JBWatenbergScality:
```suggestion
               {...register('name'), onChange: (e) => {register('name').onChange(e}; clearServerError(e); }
```

JBWatenbergScality (changes requested):

JBWatenbergScality (approved):

PR #398, opened 2022-07-06
ZKUI-215: Initializing name field in Update IAM user page

ognyan-kostadinov:


JBWatenbergScality (commented):
The change is good, however can you try to cover it with a test ?

JBWatenbergScality (approved):
LGTM test will come in a next PR

PR #401, opened 2022-07-06
ZKUI-200: Showing expiration date for current objects with active expiration flow

ognyan-kostadinov:


src/react/reducers/s3.ts
```diff
@@ -161,6 +161,15 @@ const convertToFormTags = (tags: TagSet): Tags =>
     value: t.Value,
   }));
 
+const parseExpirationDate = (expiresOn: string | undefined): Date | null => {
+  if (!expiresOn) {
+    return null;
+  }
+
+  const stripped = expiresOn.replaceAll('"', '').replaceAll('expiry-date=', '');
+  return new Date(stripped.slice(0, stripped.indexOf(', rule-id=')));
```
JBWatenbergScality: This approach is going to break if for some reason the expiry-date and rule-id get reversed (whereas it would technically still be a valid value). Maybe we should take a more resilient approach. 

For example, below regex would extract the expiry-date in the first capture group. Can be easily enhanced by naming the capture group
```suggestion
    expiresOn.exec(/expiry-date="([^"]+)"/)
```

JBWatenbergScality (approved):

PR #409, opened 2022-07-12
ZKUI-221: Fixed transition to data consumer from storage account owner

ognyan-kostadinov:


src/react/ui-elements/ReauthDialog.tsx
```diff
@@ -36,19 +48,22 @@ const ReauthDialog = () => {
...
             />
           )}
-          <Button
-            variant="primary"
-            onClick={() => window.location.reload()}
-            label="Reload"
-          />
+          <DataServiceRoleProvider>
+            <AccountRoleSelectButtonAndModal
+              bigButton
```
ChengYanJin, line 52: I would change the label to `Switch Account` in the modal.. instead of displaying the current account name.

src/react/account/AccountRoleSelectButtonAndModal.tsx
```diff
@@ -14,17 +19,38 @@ import { getRoleArnStored, setRoleArnStored } from '../utils/localStorage';
...
   const [isModalOpen, setIsModalOpen] = useState(false);
   const accountName = account?.Name;
   const [assumedAccount, setAssumedAccount] = useState(accountName);
+  const redirectDataConsummers = useRedirectDataConsumers();
+
+  useEffect(() => {
+    if (location.search === '?assumeRole') {
+      setIsModalOpen(true);
+    }
+  }, [location.search]);
```
ChengYanJin: I do believe it's not necessary to add this query param.
And in order to solve the double modal issue, we could use prepend to add the new DOM. Please see:

```js
  useEffect(() => {
    document.body && document.body.prepend(modalContainer.current);
    return () => {
      document.body && document.body.removeChild(modalContainer.current);
    };
  }, [modalContainer]);
```


https://github.com/scality/core-ui/blob/development/1.0/src/lib/components/modal/Modal.component.tsx#L67

src/react/utils/hooks.ts
```diff
@@ -186,22 +205,10 @@ export const useAccounts = () => {
...
-            !canAssumeAdminAccountRolesOnAnyAccount &&
-            !location.pathname.includes('bucket') &&
-            !location.pathname.includes('workflows') &&
-            !userGroups.includes('StorageManager')
-          ) {
-            history.replace('/buckets');
+          if (accountsWithRoles?.entries) {
+            accountsWithRoles.find((accountsWithRoles) => {
+              redirectDataConsumers(accountsWithRoles.Roles);
+            });
```
ChengYanJin: should be `forEach` to loop through all the elements

src/react/account/AccountRoleSelectButtonAndModal.tsx
```diff
@@ -42,6 +68,10 @@ function AccountRoleSelectButtonAndModal() {
 
   const handleClose = () => {
     setIsModalOpen(false);
+
+    if (location.search === '?assumeRole') {
+      history.push(location.pathname);
+    }
```
ChengYanJin: could be removed

src/react/account/AccountRoleSelectButtonAndModal.tsx
```diff
@@ -51,13 +81,15 @@ function AccountRoleSelectButtonAndModal() {
           icon={<i className="fas fa-arrow-right"></i>}
           variant="primary"
           onClick={() => {
+            const parsedArn = regexArn.exec(assumedRoleArn);
+            const roleName = parsedArn?.groups?.name || '';
             setRoleArnStored(assumedRoleArn);
             selectAccountAndRoleRedirectTo(
               path,
```
ChengYanJin: I think we have a small issue with the path here when we switch from `storage account owner` to `data consumer` in the access denied situation, the history will be pushed to '/'

```js         
selectAccountAndRoleRedirectTo(
      location.pathname + location.search,
      assumedAccount,
      assumedRoleArn,
);
 ```

File: src/react/account/AccountRoleSelectButtonAndModal.tsx [hunk 6]
```diff
@@ -144,8 +176,18 @@ function AccountRoleSelectButtonAndModal() {
...
         variant="primary"
-        onClick={() => setIsModalOpen(true)}
+        onClick={() => {
+          onClick && onClick();
+          if (
+            !location.search ||
+            location.search.indexOf('?assumeRole') === -1
+          ) {
+            history.push('?assumeRole');
+          }
```
ChengYanJin: could be removed

src/react/account/AccountRoleSelectButtonAndModal.tsx
ChengYanJin: Another behavior is missing today is:
We should do a `window. reload()` after switching roles/accounts to make sure the user still has the permission to see the previous page before opening the Select Account Modal.

The action is a bit heavy but I think it makes sense. we can try to improve it later on when we migrate listBucket/listObject to react-query.

ChengYanJin (changes requested):

src/react/account/AccountRoleSelectButtonAndModal.tsx
ognyan-kostadinov: Hm no, if that was the case it would've been working already with no intervention needed :) The problem is that when AccountRoleSelectButtonAndModal is rendered it's inside ReauthDialog. The moment in which ReauthDialog is closed, it is unmounted, thus unmounting AccountRoleSelectButtonAndModal as well. So the reason I'm doing this roundabout approach is to trigger an opening of the modal

ognyan-kostadinov: What this useEffect in core-ui's Modal component is doing is to target something outside the React Root node, to dump the DOM there, as we need it there to create the overlay which covers the entire window. But it's still managed by React and react doesn't allow you to interact with components which are not mounted. It's exactly the same reason for which ReauthDialog is pushed so far up the component hierarchy - to be always mounted

ognyan-kostadinov: There's another caveat to the current approach though - if we're at a page where there's no `<AccountRoleSelectButtonAndModal />`.While working on the issue and testing various situations - it didn't happen, but it's possible that I missed some screen in which `<ReauthDialog />` shows up and the page under the modal doesn't have `<AccountRoleSelectButtonAndModal />` . A solution to this would be to redirect to a certain page - for example `/accounts?assumeRole` but I'm not sure which page would be a good candidate for that

ChengYanJin (approved):
LGTM!

PR #411, opened 2022-07-14
ZKUI-223: Indicating that deletion is in progress in DeleteConfirmation modal

ognyan-kostadinov:


src/react/ui-elements/DeleteConfirmation.tsx
```diff
@@ -21,10 +28,12 @@ const DeleteConfirmation = ({ approve, cancel, show, titleText }: Props) => {
         <div>
           <Button variant="outline" onClick={cancel} label="Cancel" />
           <Button
+            disabled={loading}
             className="delete-confirmation-delete-button"
             variant="danger"
             onClick={() => approve()}
-            label="Delete"
+            icon={loading && <Loader size="larger" />}
+            label={loading || 'Delete'}
```
ChengYanJin: I think we can keep the Delete label while the deletion action is in progress. 
```suggestion
            label={'Delete'}
```
Or
```suggestion
            label={loading? 'Loading' : 'Delete'}
```

ognyan-kostadinov: You're right, that's a good idea! My concern with changing the button's content was that the width was dynamically changing along it, making it not so visually pleasing, but yeah, given that I've fixed it to a specific one(using rem of course) I can increase it a bit to have space for the spinner as such:  "(spinner) Delete"

ChengYanJin: Yeah. Actually, in this case, we could have an icon for idle status as well(a trash button), so the width won't change too much. But OK, I think it's good to be like this for the moment.

ChengYanJin (approved):
LGTM!

PR #419, opened 2022-07-20
ZKUI-225: Add transition table

ChengYanJin:
image
image

TODO: Create a function to return the type of workflow

### DIFF
--- a/src/react/workflow/TransitionTable.tsx
+++ b/src/react/workflow/TransitionTable.tsx
@@ -1,0 +1,115 @@
+import { Banner } from '@scality/core-ui';
+import { Box, Table } from '@scality/core-ui/dist/next';
+import { useTheme } from 'styled-components';
+import { BucketWorkflowTransitionV2 } from '../../js/managementClient/api';
+import { APIWorkflows } from '../../types/workflow';
+import { useWorkflows } from './Workflows';
+
+type Props = {
+  bucketName: string;
+  applyToVersion: BucketWorkflowTransitionV2.ApplyToVersionEnum;
+  objectKeyPrefix: string;
+  objectTags: [];
+  triggerDelayDays?: string;
+  locationName?: string;
+};
+
+const TransitionTable = ({
+  bucketName,
+  applyToVersion,
+  objectKeyPrefix,
+  objectTags,
+  triggerDelayDays,
+  locationName,
+}: Props) => {
+  const theme = useTheme();
+  const { data: transitions, status } = useWorkflows(
+    (workflows: APIWorkflows) => {
+      return workflows.filter((wf) => wf.transition).map((wf) => wf.transition);
+    },
+    {
+      objectKeyPrefix: objectKeyPrefix,
+      objectTags: objectTags,
+    },
+  );
+
+  const transitionList =
+    transitions?.filter(
+      (ts: BucketWorkflowTransitionV2) =>
+        ts.applyToVersion === applyToVersion && ts.bucketName === bucketName,
+    ) || [];
+
+  const isTransitionWithDate = transitionList?.find(
+    (ts: BucketWorkflowTransitionV2) => ts.triggerDelayDate,
+  );
+
+  // Manually add the workflow description
+  const transitionWithDesciption = transitionList.map(
+    (ts: BucketWorkflowTransitionV2) => {
+      return {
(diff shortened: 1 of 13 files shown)


src/react/workflow/TransitionTable.tsx
```diff
@@ -0,0 +1,149 @@
...
+    transitionList.find(
+      (ts: BucketWorkflowTransitionV2) => ts.triggerDelayDate,
+    );
+
+  // Manually add the workflow description
+  const transitionWithDesciption =
+    (status === 'success' &&
+      transitionList.map((ts: BucketWorkflowTransitionV2) => {
+        return {
+          description: `All current objects order than ${ts.triggerDelayDays} day(s) will transition to ${ts.locationName}`,
```
JBWatenbergScality:
```suggestion
          description: `All current objects older than ${ts.triggerDelayDays} day(s) will transition to ${ts.locationName}`,
```

JBWatenbergScality: nit: we can make it so that the `s` of `day(s)` is added only when triggerDelayDays is higher than 1

src/react/workflow/TransitionTable.tsx
```diff
@@ -0,0 +1,149 @@
...
+          ...ts,
+        };
+      })) ||
+    [];
+
+  if (triggerDelayDays && locationName) {
+    transitionWithDesciption.push({
+      triggerDelayDays,
+      locationName,
+      description: `Objects older than ${triggerDelayDays} day(s) will transition to ${locationName} `,
```
JBWatenbergScality: Same here

src/react/workflow/TransitionTable.tsx
```diff
@@ -0,0 +1,149 @@
...
+    });
+  }
+
+  const transitionWithDate = `A Transition Workflow was set up based on a specific date. We can't guarantee you the Workflow would be consistent.`;
```
JBWatenbergScality:
```suggestion
  const transitionWithDate = `A Transition Workflow was set up based on a specific date. We can't guarantee you the Workflow will be consistent.`;
```

src/react/workflow/Workflows.tsx
```diff
@@ -87,7 +104,7 @@ export default function Workflows() {
   );
 
   const select = (workflows: APIWorkflows) => makeWorkflows(workflows);
-  const workflowListDataQuery = useWorkflows(select);
+  const workflowListDataQuery = useWorkflowsWithSelect(select);
```
JBWatenbergScality: nit: select can be inlined here
```suggestion
  const workflowListDataQuery = useWorkflowsWithSelect(makeWorkflows);
```

src/react/workflow/utils.tsx
```diff
@@ -327,3 +327,38 @@ export const hasUniqueKeys = (value: Array<Tag>, helper: CustomHelpers) => {
...
+  return transitions
+    .filter((ts: BucketWorkflowTransitionV2) => {
+      if (filters.objectKeyPrefix) {
+        return ts.filter?.objectKeyPrefix === filters.objectKeyPrefix;
+      }
+      return true;
+    })
+    .filter((ts: BucketWorkflowTransitionV2) =>
+      hasIdenticalTags(ts.filter?.objectTags || [], sanitizedTags),
+    );
```
JBWatenbergScality: nit: those 2 filters could be grouped to avoid 1 iteration over workflows.

src/react/workflow/TransitionTable.tsx
```diff
@@ -67,11 +74,15 @@ const TransitionTable = ({
     transitionWithDesciption.push({
       triggerDelayDays,
       locationName,
-      description: `Objects older than ${triggerDelayDays} day(s) will transition to ${locationName} `,
+      description: `Objects older than ${triggerDelayDays} day${
+        typeof triggerDelayDays === 'string' && parseInt(triggerDelayDays) > 1
```
JBWatenbergScality: parseInt(triggerDelayDays, 10) // because js and browser implementation... if triggerDelayDays = '08' => base 8

JBWatenbergScality (approved):

src/react/workflow/TransitionTable.tsx
JBWatenbergScality: ref: https://stackoverflow.com/questions/8763396/javascript-parseint-with-leading-zeros

PR #420, opened 2022-07-21
ZKUI-228: Added a clarification message about the consequences of deleting objects and versions

ognyan-kostadinov:


ognyan-kostadinov:
multiple-objects
multiple-versions
one-object
one-version

Cuervino:
2 comments: 
- could you add "dot" at the end of the notification sentence (for instance: 
`The selected versions will be permanently removed from the Bucket.` 
)
- The font in the notification seems to be bold, where it would be better to have a TextBasic style. Maybe it's related to the Core-ui component?
image

Cuervino:
Additional question: did you check the new behaviour against "Object-locked" deletion modal ?

ChengYanJin (changes requested):
Hello @ognyan-kostadinov, 
Before checking the code, there are a few cases we should double-check with @Cuervino about the behavior. 

1. the delete marker only applies to the buckets which are **versioning**.  If the buckets are not versioning enabled, when we delete an object, it will be permanently deleted.
2. for the object with **legal hold** enabled, I think it's better to NOT display such notification (+ confirm checkbox). 
3. In the modal, if we remove all the objects, the banner should disappear as well.

ChengYanJin (approved):
Let's get it merged!
For the suspended versioning bucket, we will check it later and create a ticket if it's needed.

PR #421, opened 2022-07-21
ZKUI-230: Set minimum height for scrollbar thumb

ognyan-kostadinov:


ChengYanJin (commented):
Hi @ognyan-kostadinov,

The fix should work, but just FYI, we have ScrollbarWrapper in `core-ui` as well, can you check if we can make the change there and apply it in `zenko-ui` as well? 

Thanks!

ognyan-kostadinov:
@ChengYanJin I'm not sure if a fix there is even needed. The object table (in `ObjectListTable`) doesn't use core-ui's table, it uses `FixedSizeList` from `react-window`. And the `ScrollbarWrapper` here doesn't use core-ui's ScrollbarWrapper. I'll check if there's an issue in the storybook with very long tables in small container, but even if there is - it won't fix the problem in zenko-ui, and can be treated as a separate issue

ChengYanJin (approved):
LGTM!

PR #426, opened 2022-07-29
ZKUI-238: Fix the bug of display transition for replication

ChengYanJin:
image


MonPote (approved):
We should create a function or something to determine the type of workflow.
The double ternary is not great IMO

PR #428, opened 2022-08-01
ZKUI-237: Displaying warning when trying to delete a bucket with attached workflow

ognyan-kostadinov:


src/react/databrowser/buckets/details/Overview.tsx
```diff
@@ -56,26 +59,22 @@ function getDefaultBucketRetention(bucketInfo: BucketInfo): string {
...
 }
 
+const StyledLink = styled(Link)`
+  color: ${(props) => props.theme.brand.textLink};
+  text-decoration: none;
+  &:hover {
+    text-decoration: underline;
+  }
+`;
+
```
ChengYanJin: there is an existing styled link in src/react/ui-elements, please check if you could reuse the style.

src/react/databrowser/buckets/details/Overview.tsx
```diff
@@ -97,14 +96,43 @@ function Overview({ bucket, ingestionStates }: Props) {
...
+      >
+        delete
+      </StyledLink>{' '}
+      {count > 1 ? 'them' : 'it'} first.
+    </div>
+  );
+
+  const handleDeleteClick = (
+    bucketName: string,
+    replicationStreams: Array<Replication>,
```
ChengYanJin: actually we have three types of workflows: replication, expiration and transition. so we should check all of them.

src/react/databrowser/buckets/details/Overview.tsx
```diff
@@ -97,14 +96,43 @@ function Overview({ bucket, ingestionStates }: Props) {
   const workflowsQuery = useWorkflows();
   const features = useSelector((state: AppState) => state.auth.config.features);
   const { account } = useCurrentAccount();
+  const [isErrorModalOpen, setIsErrorModalOpen] = useState(false);
+  const [attachedWorkflowsCount, setAttachedWorkflowsCount] = useState(0);
```
ChengYanJin: Not sure we need a state for this count. I think we could get the number the workflow directly given the bucket name.

src/react/ui-elements/ErrorHandlerModal.tsx
```diff
@@ -25,15 +25,27 @@ const ErrorHandlerModal = () => {
...
   return (
-    <Modal
-      close={close}
-      footer={<Button variant="primary" onClick={close} label="Close" />}
-      isOpen={true}
-      title="Error"
-    >
-      {errorMessage}
-    </Modal>
+    <DumbErrorModal errorMessage={errorMessage} isOpen={true} close={close} />
```
ChengYanJin, line 28: If we set `isOpen={true}` does it mean the Modal will be already open?

ChengYanJin (changes requested):

src/react/ui-elements/ErrorHandlerModal.tsx
ognyan-kostadinov: It was like that before, I just extracted the JSX into a separate component in order to reuse it without the redux-store logic :) And to answer your question - no, it's not going to be always open, because of this code couple of lines before (again, I haven't written that - it was already there) :


> const showError = useSelector(
>    (state: AppState) =>
>      !!state.uiErrors.errorMsg && state.uiErrors.errorType === 'byModal',
>  );
...
if (!showError) {
    return null;
}

src/react/databrowser/buckets/details/Overview.tsx
ognyan-kostadinov: @Cuervino The pre-existing logic for preventing deletion of a bucket was specifically (and only) for attached replication workflows. Do you want me to keep it as it is, or do you want to check for expiration and transition workflows as well?

src/react/databrowser/buckets/details/Overview.tsx
ognyan-kostadinov: Yes, I copied the styles from CellLink, I'll reuse it but we have to keep in mind that if we update it in a way which makes the styling more appropriate for table elements, it might not look good in the modal. There's NameLink which is similar but it has padding right and it's styled div and yeah, using divs for links is not ideal

JBWatenbergScality: ` divs for links is not ideal`: worse than that it is not accessible !

src/react/databrowser/buckets/details/Overview.tsx
JBWatenbergScality: I think we should prevent the deletion of buckets if workflows are attached for them as well, this is something we missed before :/

src/react/databrowser/buckets/details/Overview.tsx
```diff
@@ -97,15 +136,27 @@ function Overview({ bucket, ingestionStates }: Props) {
   const workflowsQuery = useWorkflows();
```
ChengYanJin: actually we can pass `filters` which is a list of bucket name to filter the workflows directly.
```suggestion
  const workflowsQuery = useWorkflows([bucket.Name]);
```

src/react/databrowser/buckets/details/Overview.tsx
```diff
@@ -97,15 +136,27 @@ function Overview({ bucket, ingestionStates }: Props) {
...
+  const [isErrorModalOpen, setIsErrorModalOpen] = useState(false);
   useEffect(() => {
     dispatch(getBucketInfo(bucket.Name));
   }, [dispatch, bucket.Name]);
 
+  const workflows = workflowsQuery.data || {};
+  const attachedWorkflowsCount = React.useMemo(
+    () => getAttachedWorkflowsCount(bucket.Name, workflows),
+    [bucket.Name, workflows],
+  );
```
ChengYanJin:
```suggestion
  const attachedWorkflowsCount = workflows.expirations.length + workflows.replications.length + workflows.transitions.length
```

src/react/databrowser/buckets/details/Overview.tsx
```diff
@@ -94,18 +97,37 @@ function Overview({ bucket, ingestionStates }: Props) {
...
+  if (workflows && workflows.expirations) {
+    attachedWorkflowsCount += workflows.expirations.length;
+  }
+  if (workflows && workflows.replications) {
+    attachedWorkflowsCount += workflows.replications.length;
+  }
+  if (workflows && workflows.transitions) {
+    attachedWorkflowsCount += workflows.transitions.length;
+  }
+
```
ChengYanJin: just a suggestion to simplify the code a bit ;) 
```suggestion
  const attachedWorkflowsCount =
    (workflows?.expirations?.length || 0) +
    (workflows?.replications.length || 0) +
    (workflows?.transitions.length || 0);

```

ognyan-kostadinov: Yey, TS goodies come to the rescue :)

ChengYanJin (approved):
LGTM!

PR #436, opened 2022-08-04
ZKUI-247: Improve transition table in edition mode

ChengYanJin:
We have exactly the same 2 workflow in the edition mode. So filter out the with the current workflowId in the URL.

### DIFF
--- a/src/react/workflow/TransitionTable.tsx
+++ b/src/react/workflow/TransitionTable.tsx
@@ -25,10 +25,15 @@ const TransitionTable = ({
   triggerDelayDays,
   locationName,
 }: Props) => {
-  const theme = useTheme();
-  const isWorkflowCreationPage = useRouteMatch(
-    '/accounts/:accountName/workflows/create-workflow',
+  type MatchParams = {
+    accountName: string;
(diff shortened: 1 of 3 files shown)


src/react/workflow/TransitionTable.tsx
```diff
@@ -26,9 +26,10 @@ const TransitionTable = ({
   locationName,
 }: Props) => {
   const theme = useTheme();
-  const isWorkflowCreationPage = useRouteMatch(
-    '/accounts/:accountName/workflows/create-workflow',
+  const { params } = useRouteMatch(
+    '/accounts/:accountName/workflows/:workflowId',
   );
+  const transitionId = params.workflowId.split('transition-')[1] || '';
```
MonPote: I think adding a bit of typing is possible here : 
```typescript
  type MatchParams = {
    accountName: string;
    workflowId: string;
  };
  const match = useRouteMatch<MatchParams>('/accounts/:accountName/workflows/:workflowId');
  const transitionId = match?.params.workflowId.split('transition-')[1] ?? '';
```

PR #441, opened 2022-08-18
ZKUI-248: Add restore status in object property - Temperature

ChengYanJin:
Note that the screenshot is not a real case, because amazon-s3 is not a cold location.
image

### DIFF
--- a/src/react/databrowser/objects/details/Properties.tsx
+++ b/src/react/databrowser/objects/details/Properties.tsx
@@ -3,7 +3,7 @@ import Table, * as T from '../../../ui-elements/TableKeyValue2';
 import { Clipboard } from '../../../ui-elements/Clipboard';
 import MiddleEllipsis from '../../../ui-elements/MiddleEllipsis';
 import type { ObjectMetadata } from '../../../../types/s3';
-import { PrettyBytes, Toggle } from '@scality/core-ui';
+import { PrettyBytes, SecondaryText, Toggle } from '@scality/core-ui';
 import { useEffect } from 'react';
 import { formatShortDate } from '../../../utils';
 import styled from 'styled-components';
@@ -16,6 +16,7 @@ import { push } from 'connected-react-router';
 import { useLocation } from 'react-router';
 import { Button } from '@scality/core-ui/dist/next';
 import ColdStorageIcon from '../../../ui-elements/ColdStorageIcon';
+import { DateTime } from 'luxon';
 
 type Props = {
   objectMetadata: ObjectMetadata;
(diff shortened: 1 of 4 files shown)


src/react/databrowser/objects/details/Properties.tsx
```diff
@@ -121,13 +122,29 @@ function Properties({ objectMetadata }: Props) {
...
+                        <SecondaryText>
+                          (
+                          {Math.floor(
+                            DateTime.fromISO(
+                              objectMetadata.restore?.expiryDate.toISOString(),
+                            )
+                              .diff(DateTime.now(), 'days')
+                              .toObject().days,
+                          )}{' '}
+                          days remaining)
```
JBWatenbergScality: nit: if a single day is remaining then `days` should be singular

src/react/reducers/s3.ts
```diff
@@ -175,6 +175,24 @@ const parseExpirationDate = (expiresOn: string | undefined): Date | null => {
...
+  const regExpExecArray = extractOngoingRequest.exec(restore);
+  if (regExpExecArray && regExpExecArray.length > 1) {
+    if (regExpExecArray[1] === 'false') {
+      return false;
+    } else if (regExpExecArray[1] === 'true') {
+      return true;
+    }
+  }
+  return null;
+};
```
JBWatenbergScality: nit: I'm not sure to see the value of differencing the output of the function when `ongoing-request ` is defined or not, hence I would just make it return a boolean and return `false` instead of `null`

src/react/databrowser/objects/details/Properties.tsx
```diff
@@ -121,13 +122,29 @@ function Properties({ objectMetadata }: Props) {
...
+                    )}
+                    {objectMetadata.restore?.expiryDate && (
+                      <span>
+                        <i className="fas fa-stopwatch" />
+                        Restored{' '}
+                        <SecondaryText>
+                          (
+                          {Math.floor(
+                            DateTime.fromISO(
+                              objectMetadata.restore?.expiryDate.toISOString(),
```
JBWatenbergScality: nit:  this optional chaining shouldn't be necessary as it is handled by above condition
```suggestion
                              objectMetadata.restore.expiryDate.toISOString(),
```

src/react/databrowser/objects/details/Properties.tsx
```diff
@@ -55,6 +56,12 @@ function Properties({ objectMetadata }: Props) {
       dispatch(push(`${pathname}?${query.toString()}`));
     }
   }, [objectMetadata.versionId]);
+  // Display the remaining days for the restored object from a cold location
+  const restoreExpiryDays: number = Math.floor(
+    DateTime.fromISO(objectMetadata.restore?.expiryDate?.toISOString())
```
JBWatenbergScality: this may raise an error when expiryDate is not defined

ChengYanJin: Oh. yes. I will perform a check. 
Thank for the review ;)

JBWatenbergScality (approved):

PR #443, opened 2022-08-31
ZKUI-249: Replaced old tables with core-ui's TableV2

ognyan-kostadinov:


src/react/databrowser/objects/ObjectListTable.tsx
```diff
@@ -267,117 +206,34 @@ export default function ObjectListTable({
...
-                    }}
-                  >
-                    {MemoRow}
-                  </FixedSizeList>
-                )}
-              </InfiniteLoader>
-            )}
-          </AutoSizer>
-        </T.BodyWindowing>
+    <T.ContainerWithSubHeader style={{ height: '500px' }}>
```
ognyan-kostadinov: Remove fixed height

src/react/databrowser/buckets/BucketList.tsx
```diff
@@ -147,67 +148,31 @@ export default function BucketList({
...
-                  )}
-                >
-                  {MemoRow}
-                </FixedSizeList>
-              )}
-            </AutoSizer>
-          </T.BodyWindowing>
+        <Table
+          columns={columns}
+          data={bucketsImmutableList.toJS()}
```
ognyan-kostadinov: Use the already memoized buckets

src/react/databrowser/objects/ObjectListTable.tsx
```diff
@@ -267,117 +206,34 @@ export default function ObjectListTable({
...
+    bucketName,
+    dispatch,
+    handleCellClicked,
+    isTableScrollbarVisible,
+    isToggledFull,
+    isVersioningType,
+    tableWidth,
+  ]);
+
+  const objectsJS = useMemo(() => objects.toJS(), [objects.size]);
```
ognyan-kostadinov: Figure out a better name

src/react/account/details/properties/AccountKeys.tsx
```diff
@@ -83,25 +92,41 @@ function AccountKeys({ account }: Props) {
       {
         Header: 'Created On',
         accessor: 'created_at',
-        headerStyle: {
+        cellStyle: {
+          flex: '0.25',
           textAlign: 'right',
+          marginRight: '1rem',
```
ognyan-kostadinov, line 98: Pixels/rem?

src/react/databrowser/buckets/BucketList.tsx
```diff
@@ -56,9 +48,18 @@ export default function BucketList({
...
+              onClick={(event) => event.stopPropagation()}
+              to={`/accounts/${accountName}/buckets/${name}/objects`}
+            >
+              {name}
+            </T.CellLink>
           );
         },
+        cellStyle: {
+          flex: '1',
+          paddingLeft: '1rem',
```
ognyan-kostadinov, line 62: Pixels/rem?

src/react/databrowser/buckets/BucketList.tsx
```diff
@@ -133,7 +133,8 @@ export default function BucketList({
           <T.SearchInput
             disableToggle={true}
             placeholder="Search by Bucket Name"
-            onChange={(e) => setFilter('Name', e.target.value)}
+            // onChange={(e) => setFilter('Name', e.target.value)}
```
ognyan-kostadinov: Remove that

src/react/account/__tests__/AccountDetails.test.tsx
```diff
@@ -50,8 +52,9 @@ describe('AccountDetails', () => {
         },
       },
     );
-    expect(component.find(CustomTabs)).toHaveLength(1);
+
+    expect(component.getByRole('tablist')).toBeInTheDocument();
     // warning of account access key table
-    expect(component.find(Warning)).toHaveLength(1);
+    expect(component.getByTestId('warning')).toBeInTheDocument();
```
JBWatenbergScality: Just to remind you about theses undesirable testids 😉

src/react/account/details/properties/AccountKeys.tsx
```diff
@@ -83,25 +92,41 @@ function AccountKeys({ account }: Props) {
...
+            return -1;
+          }
+
+          return 0;
         },
 
-        width: 80,
+        Cell({ value: created_at }: { value: Date }) {
+          return formatShortDate(new Date(created_at));
+        },
```
JBWatenbergScality: Are you sure about this typing ? Because if `value/created_at` is already a `Date` what's the point of calling `new Date(created_at)`.

nit: why using snake case here for `created_at` ?

JBWatenbergScality: Oh I see it is defined using snake case in the data passed to the table ...

src/react/account/details/properties/AccountKeys.tsx
```diff
@@ -133,125 +158,64 @@ function AccountKeys({ account }: Props) {
...
+          <Table
+            columns={columns}
+            data={accessKeys}
+            defaultSortingKey="created_at"
+          >
+            <Table.SingleSelectableContent
+              rowHeight="h40"
+              separationLineVariant="backgroundLevel3"
+              backgroundVariant="backgroundLevel1"
+              onRowSelected={() => null}
```
JBWatenbergScality: Likely if we don't need this we should make it optional in core ui if it is not already and drop the prop here

src/react/account/details/properties/AccountKeys.tsx
```diff
@@ -133,125 +158,64 @@ function AccountKeys({ account }: Props) {
...
+              selectedId
+            />
+          </Table>
+        ) : (
+          <Warning
+            centered={true}
```
JBWatenbergScality:
```suggestion
            centered
```

src/react/databrowser/buckets/BucketList.tsx
```diff
@@ -30,24 +28,18 @@ type Props = {
...
-      e.stopPropagation();
-      dispatch(push(`/accounts/${accountName}/buckets/${name}/objects`));
-    },
-    [dispatch],
-  );
   const features = useSelector((state: AppState) => state.auth.config.features);
+  const query = useQueryParams();
+  const { account } = useCurrentAccount();
+  const tabName = query.get('tab');
+  const [bucketNameFilter, setBucketNameFilter] = useState();
```
JBWatenbergScality:
```suggestion
  const [bucketNameFilter, setBucketNameFilter] = useState<PLEASE_TYPE_THIS>(SHALL_WE_HAVE_AN_INIT_VALUE_HERE);
```

src/react/databrowser/buckets/BucketList.tsx
JBWatenbergScality: For pixel/rem consideration we will check that later on

src/react/databrowser/buckets/BucketList.tsx
```diff
@@ -81,50 +85,46 @@ export default function BucketList({
...
+    () => bucketsImmutableList.toJS(),
+    [bucketsImmutableList.size],
+  ) as S3Bucket[];
+  const selectedId = useMemo(() => {
+    if (buckets) {
+      for (let i = 0; i < buckets.length; i += 1) {
+        if (buckets[i].Name === selectedBucketName) {
+          return i;
+        }
+      }
```
JBWatenbergScality: this can probably be replaced by a `findIndex`. Also I'm not very sure that returning void when buckets  is undefined is a good thing. We should be explicit and return null I guess

src/react/databrowser/objects/ObjectListTable.tsx
JBWatenbergScality: I would suggest objectList

src/react/endpoint/EndpointList.tsx
```diff
@@ -38,28 +34,35 @@ type Props = {
...
-      return;
-    }
-
-    dispatch(deleteEndpoint(showDelete));
-  };
-
-  const handleDeleteCancel = () => {
-    dispatch(closeEndpointDeleteDialog());
-  };
+  const [hostnameFilter, setHostNameFilter] = useState();
```
JBWatenbergScality: Nit: just for consistency 
```suggestion
  const [hostnameFilter, setHostnameFilter] = useState();
```

src/react/endpoint/EndpointList.tsx
```diff
@@ -38,28 +34,35 @@ type Props = {
...
+   *   specified in the custom render (Cell function)
+   *   Currently the API returns some objects without a `isBuiltin` property causing this undesired behavior
+   */
+  const strictSchemaEndpoints = useMemo(
+    () =>
+      endpoints.map((endpoint) => ({
+        ...endpoint,
+        isBuiltin: Boolean(endpoint.isBuiltin),
+      })),
+    [endpoints],
```
JBWatenbergScality, line 54: Is endpoints stable across re-renders ?

src/react/endpoint/EndpointList.tsx
```diff
@@ -157,54 +158,19 @@ function EndpointList({ endpoints, locations }: Props) {
...
-                  {MemoRow}
-                </FixedSizeList>
-              )}
-            </AutoSizer>
-          </T.BodyWindowing>
+        <Table
+          columns={columns}
+          data={strictSchemaEndpoints}
+          defaultSortingKey="isBuiltin"
+          allFilters={[{ id: 'hostname', value: hostnameFilter }]}
```
JBWatenbergScality, line 169: I guess now you want `hostnameFilter` to be undefined by default ?

src/react/endpoint/EndpointList.tsx
```diff
@@ -157,54 +158,19 @@ function EndpointList({ endpoints, locations }: Props) {
...
+        >
+          <Table.SingleSelectableContent
+            rowHeight="h40"
+            selectedId
+            onRowSelected={() => null}
```
JBWatenbergScality: Same as above we could drop this props

src/react/account/details/properties/AccountKeys.tsx
ognyan-kostadinov: `onRowSelected` is optional already. This is kind of a hack on my side (onRowSelected being a no-op and selectedId being defined but empty) to get the table to change the background of rows when hovering over them, but of course they're not actually selectable - nothing happens when you click on them. I did it because the previous table had that functionality, I'm okay to drop it if we don't need it. Or introduce a new prop `hoverable` which does only that. In some cases, when the action buttons are at the distant right end of the table and you're unsure which row you're targeting because there's a lot of empty space between the last column and action buttons, it's very useful to have that visual aid (for example I'm a bit astigmatic on the right eye and it affects me)

src/react/endpoint/EndpointList.tsx
ognyan-kostadinov: It's as stable as useSelector provides it, no transformations are performed to it in the parent container

src/react/account/details/properties/AccountKeys.tsx
ognyan-kostadinov: Well, apparently `value` is not a Date, it's string, because when I removed it some tests started failing :D

ognyan-kostadinov: The typing of `value` was wrong, it's a string and converting it to date is needed

src/react/account/details/properties/AccountKeys.tsx
ChengYanJin: To me, the background color of the row the this accountKey table doesn't need to change, since it's not really actionable.  So I think we can safely remove the ```onRowSelected={() => null}```

src/react/databrowser/objects/ObjectListTable.tsx
```diff
@@ -110,59 +87,23 @@ export default function ObjectListTable({
...
+        },
 
         Cell({ row: { original } }: CellProps) {
           if (original.isFolder) {
             return (
               <span>
                 <Icon className="far fa-folder"></Icon>
                 <T.CellClick
-                  onClick={handleCellClicked(bucketName, original.key)}
+                // onClick={handleCellClicked(bucketName, original.key)}
```
ChengYanJin: Should uncomment the onClick function no?

ChengYanJin (changes requested):
I tested your PR by running the UI locally ;) 
We forgot to add the prefix in the URL when we select an object.

src/react/endpoint/EndpointList.tsx
ChengYanJin: Yes. I think it's the same as accesskey table, it's not necessary to change the background color when we hover it. So we could drop this.

src/react/account/__tests__/AccountDetails.test.tsx
```diff
@@ -1,9 +1,9 @@
 import router, { MemoryRouter } from 'react-router';
 import AccountDetails from '../AccountDetails';
-import { CustomTabs } from '../../ui-elements/Tabs';
 import React from 'react';
-import { Warning } from '../../ui-elements/Warning';
-import { reduxMount } from '../../utils/test';
+import { reduxRender } from '../../utils/test';
+import { screen } from '@testing-library/react';
```
ChengYanJin, line 5: could be removed ;)

src/react/databrowser/objects/ObjectListTable.tsx
```diff
@@ -50,30 +97,20 @@ export default function ObjectListTable({
   prefixWithSlash,
 }: Props) {
   const { accountName } = useParams<{ accountName: string }>();
-
-  const [isTableScrollbarVisible, setIsTableScrollbarVisible] = useState(false);
-  const [tableWidth, setTableWidth] = useState(0);
+  const { pathname } = useLocation();
+  const query = useQueryParams();
+  const [isTableScrollbarVisible] = useState(false);
```
ChengYanJin: Since `isTableScrollbarVisible` is always false, do we still need a state or further more do we even need to declare this variable? I will let you to double check.

src/react/databrowser/objects/ObjectListTable.tsx
```diff
@@ -50,30 +97,20 @@ export default function ObjectListTable({
...
+  const [tableWidth] = useState(0);
```
ChengYanJin: Same for `tableWidth`

src/react/databrowser/objects/ObjectListTable.tsx
```diff
@@ -50,30 +97,20 @@ export default function ObjectListTable({
...
   const dispatch = useDispatch();
-  const nextMarker = useSelector(
-    (state: AppState) => state.s3.listObjectsResults.nextMarker,
-  );
-  const loading = useSelector(
-    (state: AppState) => state.networkActivity.counter > 0,
-  );
+  const { bucketName: bucketNameParam } = useParams();
```
ChengYanJin: I guess we could improve the typing by the way.
```suggestion
 const { bucketName: bucketNameParam } = useParams<{ bucketName: string }>();
```

src/react/databrowser/objects/ObjectListTable.tsx
```diff
@@ -50,30 +97,20 @@ export default function ObjectListTable({
...
+
   const locations = useSelector(
     (state: AppState) => state.configuration.latest.locations,
   );
```
ChengYanJin: typing:
```suggestion
  const locations = useSelector(
    (state: AppState) => state.configuration.latest?.locations,
  );
```

src/react/databrowser/objects/ObjectListTable.tsx
```diff
@@ -267,117 +241,124 @@ export default function ObjectListTable({
...
+    isVersioningType,
+    isVersioningDisabled,
+    tableWidth,
+  ]);
+
+  const searchInput = query.get('metadatasearch');
+  const objectList = useMemo(
+    () => objects.toJS(),
+    [objects.size],
+  ) as ObjectEntity[];
```
ChengYanJin: I think it's here that cause the issue of object list didn't get updated.

src/react/databrowser/objects/ObjectListTable.tsx
```diff
@@ -267,117 +241,124 @@ export default function ObjectListTable({
...
-                  </FixedSizeList>
-                )}
-              </InfiniteLoader>
-            )}
-          </AutoSizer>
-        </T.BodyWindowing>
+      <Table
+        columns={columns}
+        data={objectList}
+        defaultSortingKey="storageClass"
```
ChengYanJin: Not sure the default sorting key should be storagclass. Will check.

ChengYanJin (approved):
The comments I push was made before for the ObjectListTable. We can ignore it for now. 
LET'S GO!!!

PR #478, opened 2022-10-06
ZKUI 269 Fix account list user and policies regression

ghivert:
Fix the display bugs on Account User and Policies.

This removes the unneeded margins, and now use a proper stack with gap to space children.


ChengYanJin (commented):
Hello @ghivert,
image

Figma design:
image

I discovered three small discrepancy compare to our Figma design:
1. the color of the question mark icon
2. the color of the + icon 
3. the height of the Edit button is not the same with others. 


Thanks!

ChengYanJin (approved):
LGTM!

PR #480, opened 2022-10-07
ZKUI-264 Use new form components workflow

ghivert:


src/react/workflow/CreateWorkflow.tsx
```diff
@@ -297,103 +303,114 @@ const CreateWorkflow = () => {
...
+        layout={{ kind: 'page', title: 'Create New Workflow' }}
+        onSubmit={handleSubmit(onSubmit)}
+        rightActions={
+          <Stack gap="r16">
+            <Button
+              disabled={loading}
+              id="cancel-workflow-btn"
+              style={{
+                marginRight: spacing.sp24,
+              }}
```
ChengYanJin: Could be removed, gap 'r16' should be enough.

src/react/workflow/CreateWorkflow.tsx
```diff
@@ -297,103 +303,114 @@ const CreateWorkflow = () => {
...
+            direction="horizontal"
+            id="type"
+            label="Rule Type"
+            content={
+              <Controller
+                control={control}
+                name="type"
+                render={({ field: { onChange, onBlur, value: type } }) => {
+                  return (
+                    <Box style={{ width: '20.5rem' }}>
```
ChengYanJin: Lately, we introduced a size prop for SelectV2 component.  By default it takes 20.5 rem ;)

src/react/workflow/ReplicationForm.tsx
```diff
@@ -137,82 +157,57 @@ function ReplicationForm({
...
+                        </span>
+                      );
                     }
-
-                    return (
+                    return renderSource(locations)(result);
+                  }
+                  return (
+                    <Box
+                      style={{ width: '20.5rem' }}
```
ChengYanJin: could be removed

src/react/workflow/ReplicationForm.tsx
```diff
@@ -235,58 +230,46 @@ function ReplicationForm({
...
+            }
+            content={
+              <Controller
+                control={control}
+                name={`${prefix}sourcePrefix`}
+                render={({
+                  field: { onChange, onBlur, value: sourcePrefix },
+                }) => {
+                  return (
+                    <Input
```
ChengYanJin, line 256: Please use directly the Inputv2 component from core-ui. And the the size should be the default one which is 20.5rem

src/react/workflow/ReplicationForm.tsx
```diff
@@ -137,82 +157,57 @@ function ReplicationForm({
...
+          forceLabelWidth={forceLabelWidth}
+        >
+          <FormGroup
+            required
+            id="sourceBucket"
+            labelHelpTooltip={
+              isCreateMode
+                ? 'Source bucket has to be versioning enabled'
+                : undefined
+            }
```
ChengYanJin: I personally vote to put this sentence as the helper.  To be check with @Cuervino

src/react/workflow/TransitionForm.tsx
```diff
@@ -106,299 +157,221 @@ export const TransitionForm = ({
...
+            error={errors[`${prefix}filter.objectTags`]?.message}
+            helpErrorPosition="bottom"
+            content={
+              <Controller
+                name={`${prefix}filter.objectTags`}
+                control={control}
+                defaultValue={[{ key: '', value: '' }]}
+                render={({ field: { onChange, value } }) => {
+                  return (
+                    <TagsFilter
```
ChengYanJin: I suggest we refactor the TagsFiller as well with InputV2 size '1/2'

src/react/workflow/TransitionForm.tsx
```diff
@@ -106,299 +157,221 @@ export const TransitionForm = ({
...
+          forceLabelWidth={forceLabelWidth}
+        >
+          <FormGroup
+            label="Prefix"
+            id="prefix"
+            direction="horizontal"
+            error={errors[`${prefix}filter.objectKeyPrefix`]?.message}
+            helpErrorPosition="right"
+            content={
+              <Input
```
ChengYanJin, line 211: the size of the Inputv2 should be the default one 20.5rem

src/react/workflow/TransitionForm.tsx
```diff
@@ -106,299 +157,221 @@ export const TransitionForm = ({
...
+              direction="horizontal"
+              content={
+                <Controller
+                  control={control}
+                  name={`${prefix}triggerDelayDays`}
+                  render={({
+                    field: { onChange, value: triggerDelayDays },
+                  }) => {
+                    return (
+                      <Input
```
ChengYanJin: Let's try to use InputV2 component.

Just FYI: Our goal in general is to make the width of `Select` and `Input` the same in the Form.

src/react/workflow/TransitionForm.tsx
```diff
@@ -106,299 +157,221 @@ export const TransitionForm = ({
...
-                <T.ErrorContainer>
-                  <ErrorInput
-                    id="error-objectTags"
-                    error={errors[`${prefix}filter.objectTags`]?.message}
+                  <Box ml={1}>Current version</Box>
+                </F.Label>
+                <F.Label
+                  htmlFor="previous"
+                  style={{ alignItems: 'baseline', width: 'auto' }}
+                >
```
ChengYanJin, line 280: I suggest we completely drop the F style.

src/react/workflow/TransitionForm.tsx
```diff
@@ -106,299 +157,221 @@ export const TransitionForm = ({
...
-                  triggerDelayDays={triggerDelayDays}
-                  locationName={locationName}
+                    Previous version
+                  </Box>
+                  {isPreviousVersionDisabled ? (
+                    <>
+                      <IconHelp
+                        tooltipMessage="This action is disabled when source bucket is not versionned"
+                        overlayStyle={{ width: '13rem' }}
+                      />
```
ChengYanJin: I think in general when when option/toggle is disabled, it would be more obvious to use a helper text to explain the reason than use a IconHelp tooltip.

src/react/workflow/ExpirationForm.tsx
```diff
@@ -134,197 +182,185 @@ export function ExpirationForm({ bucketList, locations, prefix = '' }: Props) {
...
-              <T.Row data-testid="toggle-action-expire-current-version">
-                <T.KeyTooltip
-                  tooltipMessage={`
+        <FormSection
+          title={{ name: 'Action' }}
+          forceLabelWidth={forceLabelWidth}
+        >
+          <T.GroupContent required>
+            <T.Row data-testid="toggle-action-expire-current-version">
+              <T.KeyTooltip
```
ChengYanJin: I think we should try to get rid of using the legacy T style ;) 
And try to use `<FormGroup/>` for each of the Row/Action.

src/react/workflow/ExpirationForm.tsx
```diff
@@ -134,197 +182,185 @@ export function ExpirationForm({ bucketList, locations, prefix = '' }: Props) {
...
-                  Expire <span style={{ fontWeight: 'bold' }}>
-                    Current
-                  </span>{' '}
-                  version of objects{' '}
-                </T.KeyTooltip>
-                <T.Value
+                tooltipWidth="13rem"
+                size={60}
+              >
+                Expire <span style={{ fontWeight: 'bold' }}>Current</span>{' '}
```
ChengYanJin: We introduced the lately the Text component in core-ui.
```suggestion
                Expire <Text isEmphazed>Current</Text>{' '}
```

src/react/workflow/ExpirationForm.tsx
```diff
@@ -335,115 +371,105 @@ export function ExpirationForm({ bucketList, locations, prefix = '' }: Props) {
...
                       return (
-                        <Toggle
-                          id="currentVersionTriggerDelayDaysToggle"
-                          placeholder="currentVersionDelayDaysToggle"
-                          toggle={
-                            currentVersionTriggerDelayDays !== null &&
-                            currentVersionTriggerDelayDays !== undefined
-                          }
+                        <Input
+                          id="currentVersionT  f riggerDelayDays"
```
ChengYanJin: :)

src/react/workflow/ExpirationForm.tsx
```diff
@@ -335,115 +371,105 @@ export function ExpirationForm({ bucketList, locations, prefix = '' }: Props) {
...
+                    );
+                  }}
+                />
+
+                {!isSourceBucketVersionned ? (
+                  <IconHelp
+                    tooltipMessage="This action is disabled when source bucket is not versionned"
+                    variant="outline"
+                    overlayStyle={{ width: '13rem' }}
+                  />
```
ChengYanJin: Would be better to display as a helper text instead of within the tooltip.

src/react/workflow/ExpirationForm.tsx
```diff
@@ -134,197 +182,185 @@ export function ExpirationForm({ bucketList, locations, prefix = '' }: Props) {
...
-            <T.GroupName>Action</T.GroupName>
-            <T.GroupContent>
-              <T.Row data-testid="toggle-action-expire-current-version">
-                <T.KeyTooltip
-                  tooltipMessage={`
+        <FormSection
+          title={{ name: 'Action' }}
+          forceLabelWidth={forceLabelWidth}
+        >
+          <T.GroupContent required>
```
ChengYanJin: When the action is disabled, I think we should add `disabled` props to `<FormGroup/>` to make the action title disabled state as well.

ChengYanJin (commented):
Some of my comments might be duplicated, so here are the summary: 

- Try to use the latest Selectv2 and Inputv2 directly, and make sure they have the same width.
- Let's get rid of the legacy T, F style. the reason we build the Form component is harmonise the style. 
- When the action is disabled, I suggest to use helper text rather than IconHelp tooltip to explain the reason. 

Thanks for addressing the comments.

src/react/workflow/TransitionForm.tsx
ghivert: That's something I'd like, but we don't have radio inputs in Core-UI atm

src/react/workflow/TransitionForm.tsx
ghivert: @Cuervino ?

src/react/workflow/ExpirationForm.tsx
ghivert: @Cuervino ?

src/react/workflow/ReplicationForm.tsx
Cuervino: Perfect, sure.

src/react/workflow/ExpirationForm.tsx
Cuervino: 1. I'm not 100% sure, adding too many helper texts might be cumbersome.
2. it should be "versioned", not "versionned".

ghivert: Capture d’écran 2022-10-10 à 18 17 18

We got something like that

Cuervino: I prefer the tooltip, here we don't know with what the test is related to.

src/react/workflow/ConfigurationTab.tsx
```diff
@@ -552,93 +569,101 @@ function EditForm({
...
-          </form>
-        </FormContainer>
+            </Stack>
+          }
+        >
+          <FormSection forceLabelWidth={convertRemToPixels(10)}>
+            <FormGroup
+              required
+              label="Type"
+              disabled
```
ChengYanJin: I suggest to not disable this FormGroup, and change the lable to "Rule Type"

ChengYanJin (approved):
A small comment, sorry I forgot to mention in the first review. 
Other than this, LGTM !!! Thanks !!

PR #481, opened 2022-10-11
WIP: Use the new table component in data browser

ognyan-kostadinov:


ognyan-kostadinov:
Important notes:

Once ZKUI-249 is merged, this branch needs to be rebased on development/1.6 (or whatever version we'll be targeting then) and to exclude the ZKUI-249 commit.

We need also to extract the "toggled" behavior into a separate list, currently it resides in each object. The problem with that is that each time we toggle something, we update the objects lists, thus causing a re-render. Not only this is sub-optimal, but `react-table` (the package used in table v2 component) loses the selection once we provide new `data` prop. Re-rendering fixes the issues which we have currently when browsing into a folder, but at the price of the loss of selection.

The other known problem is that the list behaves strangely when selecting an item and enabling "Versions list". In general the versioned view is buggy, sometimes showing only the first item, so it must be inspected.

PR #494, opened 2022-10-24
Introduce the new layout

ognyan-kostadinov:
Yanjin: 
I just add some screenshots for the new layout integration:
image
image
image
image


src/react/databrowser/buckets/Buckets.tsx
```diff
@@ -71,21 +71,39 @@ export default function Buckets() {
...
+                ingestionStates={ingestionStates}
+              />
+            ),
+          }}
         />
-        <BucketDetails bucket={bucket} ingestionStates={ingestionStates} />
-      </L.Body>
-    </L.ContentContainer>
+      </AppContainer.MainContent>
+      {/* <L.Body></L.Body> */}
```
ChengYanJin:
```suggestion
```

src/react/databrowser/objects/Objects.tsx
```diff
@@ -127,6 +132,10 @@ export default function Objects() {
   // }
   return (
     <L.ContentContainer>
```
ChengYanJin: While I am seeing the changes in the ListLayoutx.ts, I am wondering if we could remove them as much as possible given we have out new Layout Component. 
For example, if we could get rid of this `<L.ContentContainer/>` ?

ChengYanJin (changes requested):
I have noticed that we lost the scroll ability in some of the pages:
- In some of the long creation forms. (Create Expiration,Transition workflow, Create a New Bucket with Object-lock Enabled)
image

- The Right Panel of Workflow list
image

I will let you investigate a bit the reason. Thanks.

src/react/ZenkoUI.tsx
```diff
@@ -48,11 +48,11 @@ function ZenkoUI() {
 
     if (isConfigLoaded) {
       return (
-        <ZenkoUIContainer>
```
ChengYanJin, line 51: the import can be removed ;)

src/react/account/AccountContent.tsx
```diff
@@ -65,13 +61,11 @@ function AccountContent() {
             <AccountHead />
           </AppContainer.OverallSummary>
           <AppContainer.MainContent background="backgroundLevel1">
-            <L.Content>
-              <AccountDetails account={account} />
-            </L.Content>
+            <AccountDetails account={account} />
```
ChengYanJin, line 70: Just wondering if we can get rid of  <L.BreadcrumbContainer> in `ListLayout5.ts`? 

Also I think the following definition are not being used anymore.

```js
export const Container = styled.div`
  display: flex;
  flex-direction: column;
  flex: 1;
`;
export const Content = styled.div`
  flex: 1;
  height: 100%;
  background-color: ${(props) => props.theme.brand.backgroundLevel3};
`;
```

src/react/account/AccountContent.tsx
```diff
@@ -65,13 +61,11 @@ function AccountContent() {
             <AccountHead />
```
ChengYanJin: Instead of using the legacy `<AccountHead>` we could try something as below:
```js
<Stack gap="r16">
  <Icon name="Account" color="infoPrimary" size="2x" withWrapper />
  <Text variant="Larger">{accountNameParam}</Text>
</Stack>
```

In general, what I'd like is to get rid of the ListLayout / ListLayout2 / ListLayout3 / ListLayout4 / ListLayout5 as much as possible since we have the new Layout component, could you check that please?

ChengYanJin (approved):
LGTM! The UI looks really great and neat now !

PR #502, opened 2022-11-04
ZKUI-288 Support for preferredReadLocation in ReplicationForm

ghivert:


src/react/workflow/ReplicationForm.tsx
```diff
@@ -386,104 +410,181 @@ const RenderDestination = ({
...
-                            const newValues = [...destinationLocations];
-                            newValues[index] = value;
-                            onChange(newValues);
-                            trigger();
-                          }}
-                          value={destLoc}
+    <FormSection
+      title={{ name: 'Destination' }}
+      forceLabelWidth={forceLabelWidth}
+    >
```
ChengYanJin, line 422: The code and logic look very good to me! 

Just 2 small comments:

1. The `Destination` Section is not optional. 
We indeed already have the logic within <Form/> component in core-ui to compute whether it's optional. 
I guess the logic doesn't meet this specific case, I'll let you to check if there is a way to quickly fix this. 

2. And by the way maybe we can increase a bit the `forceLabelWidth`, in order to put `Preferred Read Location *` in one line?

Thanks! 
image

src/react/workflow/ReplicationForm.tsx
```diff
@@ -421,6 +421,11 @@ const RenderDestination = ({
       forceLabelWidth={forceLabelWidth}
     >
       <Controller
+        // This required hack is there to indicates to FormSection to remove the
+        // (optional) text.
+        // eslint-disable-next-line
+        // @ts-ignore
+        required
```
JBWatenbergScality, line 428: I'm a bit surprised that this hack would works 🤔.

ghivert: And yet it *is* working ha ha 😁

ghivert: The idea is to provide a `required` prop into a child (any child in fact) because `FormSection` has an algorithm which is roughly:

```typescript
const children = Children.toArray(props.children)
const required = !!children.find(c => 'required' in c.props)
```

Even if the child does not use `required`, it *is there* and thus, `FormSection` will not display `(optional)`. Why wouldn't it work in your mind?

ghivert:
Boop

src/react/workflow/ReplicationForm.tsx
JBWatenbergScality: I thought we were relying on a context defined on the input! So alright !

JBWatenbergScality (approved):

PR #536, opened 2023-02-07
Always display wasabi locations, bump nginx base docker image and do not rely on external cdn for monaco js

JBWatenbergScality:


ChengYanJin (approved):
thanks for the batch fixes !

PR #549, opened 2023-03-13
Pause and Resume

ognyan-kostadinov:


src/react/actions/zenko.ts
JBWatenbergScality, line 1: I'm not sure any modification is necessary in this file.

src/react/backend/location/Locations.tsx
```diff
@@ -120,13 +127,37 @@ function Locations() {
...
+        instanceId,
+      ) as InstanceStatus;
+    },
+    onError: (error) => {
+      dispatch(handleClientError(error));
+    },
+    refetchOnMount: false,
+    refetchOnWindowFocus: false,
+  });
+
```
JBWatenbergScality: As this instanceStatus query is only used by PauseAndResume component, I would suggest to move it there, similarly as for the invalidation logic

src/react/backend/location/PauseAndResume.tsx
```diff
@@ -0,0 +1,146 @@
...
+                    'Async Metadata updates is active.'}
+                </Box>
+              ),
+              placement: 'top',
+            }}
+            label="Pause"
+            onClick={() => {
+              if (replication === 'enabled') {
+                dispatch(networkStart('Pausing Replication workflow'));
+                pauseReplicationSiteMutation.mutate(locationName);
```
JBWatenbergScality: IMO, this dispatch network start shall be handled by the mutation

src/react/backend/location/PauseAndResume.tsx
```diff
@@ -0,0 +1,146 @@
...
+              }
+              if (ingestion === 'enabled') {
+                dispatch(networkStart('Pausing Async Metadata updates'));
+                pauseIngestionSiteMutation.mutate(locationName);
```
JBWatenbergScality: IMO, this dispatch network start shall be handled by the mutation

src/react/backend/location/Locations.tsx
```diff
@@ -120,13 +127,37 @@ function Locations() {
...
+    useSelector((state: AppState) => state.instances.selectedId),
   );
+
+  const managementClient = useManagementClient();
+
+  const queryClient = useQueryClient();
+  const invalidateInstanceStatusQueryCache = async () => {
+    await queryClient.invalidateQueries(['instanceStatus', instanceId]);
+    await queryClient.refetchQueries(['instanceStatus', instanceId]);
+    dispatch(networkEnd());
```
JBWatenbergScality: This should also be called for the pause and resume mutation in case of failures

src/react/backend/location/Locations.tsx
```diff
@@ -174,18 +171,39 @@ function Locations() {
...
+              <>
+                Pausing the Workflow statuses will halt any workflow processes,
+                including asynchronous metadata updates and replication, that
+                are targeting this location.
+                <br /> <br />
+                Any new object added to the source location will be queued and
+                processed only once the Workflow processes are resumed.Name of
+                the bucket/container created in the specific location (e.g.
+                RING, Azure, AWS S3, GCP...), and where buckets attached to that
+                location will store data.
```
JBWatenbergScality:
```suggestion
                processed only once the Workflow processes are resumed.
```

src/react/backend/location/PauseAndResume.tsx
```diff
@@ -0,0 +1,178 @@
...
+export const PauseAndResume = ({ locationName }: { locationName: string }) => {
+  const dispatch = useDispatch();
+  const instanceId = notFalsyTypeGuard(
+    useSelector((state: AppState) => state.instances.selectedId),
+  );
+  const managementClient = useManagementClient();
+  const queryClient = useQueryClient();
+  const invalidateInstanceStatusQueryCache = async () => {
+    await queryClient.invalidateQueries(['instanceStatus', instanceId]);
+    await queryClient.refetchQueries(['instanceStatus', instanceId]);
```
JBWatenbergScality: Why do we need both ? I assume `await queryClient.refetchQueries(['instanceStatus', instanceId]); ` should be enough here.

src/react/backend/location/PauseAndResume.tsx
```diff
@@ -0,0 +1,178 @@
...
+      return managementClient?.getLatestInstanceStatus(
+        instanceId,
+      ) as InstanceStatus;
+    },
+    onError: (error) => {
+      dispatch(handleClientError(error));
+    },
+    refetchOnMount: false,
+    refetchOnWindowFocus: false,
+  });
```
JBWatenbergScality: I would extract this query into a queries object similarly as we do for other features. That will enable simpler reuse of the queries later on

src/react/backend/location/PauseAndResume.tsx
JBWatenbergScality: In addition once the query would be extracted we could retrieve the query id from the extracted query directly

src/react/backend/location/__tests__/PauseAndResume.test.tsx
```diff
@@ -0,0 +1,211 @@
...
+    );
+    reduxRender(<PauseAndResume locationName={locationName} />, {
+      instances: {
+        selectedId: instanceId,
+      },
+    });
+
+    await waitForElementToBeRemoved(() => screen.queryByTestId('loader'), {
+      timeout: 8000,
+    });
```
JBWatenbergScality: Can we use below option instead? 

```suggestion
    await waitForElementToBeRemoved(() => screen.queryByText('Loading'), {
      timeout: 8000,
    });
```

JBWatenbergScality (approved):

PR #552, opened 2023-03-16
ZKUI-316: Replication on non versioned bucket lacks error report

ChengYanJin:
- Disable the replication choice for the non versioned bucket. 
- Fix the bug that no error modal is display when the creation failed.
- Add two test cases to cover the above cases.


JBWatenbergScality (approved):
i-approve-gif

PR #560, opened 2023-04-05
ZKUI-327: Init the domain for accounts, buckets and locations and pensieve metric adaptor

ChengYanJin:
We created `next-architecture` folder which contains `adaptor`, `domain` and `ui`. 
Eventually we will remove this layer.


JBWatenbergScality (approved):
digibyte-dgb-gif

PR #564, opened 2023-04-12
ZKUI-331: Update pensieve api with storage consumption

ChengYanJin:


JBWatenbergScality (approved):
spongebob-thumbs-up-gif

PR #586, opened 2023-05-15
[ZKUI-349] ✨ Implement GHAS

gaspardmoindrot:


.github/workflows/security.yaml
```diff
@@ -0,0 +1,79 @@
...
+  workflow_dispatch:
+
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    permissions:
+      contents: read
+      packages: write
+    steps:
```
tcarmet: as this repo is not using any kind of internally made reuasable workflow, feel free to use the one on scality/workflows.

.github/workflows/security.yaml
```diff
@@ -0,0 +1,79 @@
...
+          cache-to: type=gha,mode=max
+
+      - name: Build and push keycloak
+        uses: docker/build-push-action@v2.7.0
+        with:
+          push: true
+          context: .github/docker/keycloakconfig
+          tags: ghcr.io/${{ github.repository }}/ci-keycloak:${{ github.sha }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
```
tcarmet: You're building this image, but you're not scanning it, why?

tcarmet (changes requested):

.github/workflows/security.yaml
gaspardmoindrot: Is it better now ?

.github/workflows/security.yaml
```diff
@@ -0,0 +1,45 @@
...
+    uses: scality/workflows/.github/workflows/trivy.yaml@v1
+    with:
+      name: zenko-ui
+
+  trivy-keycloak:
+    needs: build-keycloak
+    name: Trivy Vulnerability Scan for Code
+    uses: scality/workflows/.github/workflows/trivy.yaml@v1
+    with:
+      name: ci-keycloak
```
JBWatenbergScality: I don't think we would want to scan this image. it's only used for running automatic tests and not bundled in any release.

.github/workflows/security.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+
+jobs:
+  build-zenko-ui:
+    permissions:
+      contents: read
+      packages: write
+    uses: scality/workflows/.github/workflows/docker-build.yaml@v1
+    with:
+      name: zenko-ui
```
JBWatenbergScality, line 23: Instead of rebuilding the image, we can run the scan on the image matching the head of the branch.

JBWatenbergScality (commented):
Hello, Thanks for your contribution, few comments on my side.

JBWatenbergScality (commented):
Hello, Thanks for your contribution, few comments on my side.

.github/workflows/security.yaml
tcarmet: Agree

.github/workflows/security.yaml
tcarmet: We might be facing a race condition in this case. If we want this workflow to be independent we should be building this image.

gaspardmoindrot: You mean using the build that is created in `build-push-tests.yml`, right? In this case, indeed, we may have a race condition issue since we cannot use cross-file `needs`.

.github/workflows/security.yaml
gaspardmoindrot: Thanks, I modify this!

tcarmet (approved):

gaspardmoindrot:
`ping`

gaspardmoindrot:
`ping`

PR #590, opened 2023-05-23
ZKUI-353 - hide side bar for form pages

MonPote:


MonPote:
ping

MonPote:
Close this one because already integrated in #591

PR #607, opened 2023-06-14
ZKUI-364 - ability to delete non empty objects

hervedombya:
https://github.com/scality/zenko-ui/assets/135591453/ae4bf553-6cd5-4463-be33-ae0eea67869c


src/react/actions/s3bucket.ts
```diff
@@ -47,6 +49,19 @@ export function closeBucketDeleteDialog(): CloseBucketDeleteDialogAction {
...
+  return {
+    type: 'OPEN_BUCKET_EMPTY_DIALOG',
+    bucketName,
+  };
+}
+export function closeBucketEmptyDialog(): CloseBucketEmptyDialogAction {
+  return {
+    type: 'CLOSE_BUCKET_EMPTY_DIALOG',
+  };
+}
```
JBWatenbergScality: We aim to drop redux, for this simple state controlling the modal status I think it should be co-located with the empty button and it's modal and not deported to redux, this would ease unit testing and maintainability significantly .

src/react/ui-elements/EmptyConfirmation/EmptyConfirmationFooter.tsx
```diff
@@ -0,0 +1,41 @@
...
+      <p></p>
+      <Stack>
+        <Button
+          variant="outline"
+          onClick={cancel}
+          label={EMPTY_CANCEL_BUTTON_LABEL}
+        />
+        <Button
+          disabled={!isConfirm || loading}
+          className="delete-confirmation-delete-button"
```
JBWatenbergScality: nit: Why do we need this className ?

src/react/ui-elements/EmptyConfirmation/EmptySummary.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const deleteResultData = useCreateDeleteResultData(deleteResult);
+
+  const close = useCallback(() => dispatch(closeBucketEmptyDialog()), []);
+
+  return (
+    <Modal
+      title={SUMMARY}
+      close={onClose}
+      isOpen
+      footer={<EmptySummaryFooter {...{ close }} />}
```
JBWatenbergScality: nit:
```suggestion
      footer={<EmptySummaryFooter  close={close} } />}
```

src/react/ui-elements/EmptyConfirmation/EmptySummaryList.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const isSuccess = status === DeleteStatus.DELETED;
+
+  return (
+    <Box display="flex" gap={8} alignItems="center">
+      {isSuccess ? (
+        <Icon name={'Exclamation-circle'} color={'statusHealthy'} />
+      ) : (
+        <Icon name={'Exclamation-circle'} color={'statusCritical'} />
+      )}
+      {isSuccess ? DELETE_SUCCESS : DELETE_FAILED}
```
JBWatenbergScality: nit: We have to review thes to align them with the attachment status

src/react/ui-elements/EmptyConfirmation/EmptySummaryList.tsx
```diff
@@ -0,0 +1,89 @@
...
+    ],
+    [],
+  );
+
+export const EmptySummaryList = ({ data }: EmptySummaryListProps) => {
+  const columns = useCreateColumns();
+
+  if (!data.length) {
+    return null;
+  }
```
JBWatenbergScality: We generally prefer to use the empty table state so that the layout stay consistent between empty states and non empty states

src/react/ui-elements/EmptyConfirmation/useEmptyBucket.ts
```diff
@@ -0,0 +1,135 @@
...
+  const [delimiter, setDelimiter] = useState<DelimiterType>({});
+
+  const {
+    mutate: emptyBucket,
+    isLoading,
+    isSuccess,
+    isError,
+  } = useDeleteObjects();
+
+  const fetchObjects = async () => {
```
JBWatenbergScality: I think we can simplify this significantly thanks to our custom useAwsPaginatedResource hook. Also this implementation have flows if the modal get closed while it is performing the action as it will ends up in trying to update some state on an unmounted component. React Query have builtin ways to handle those things, we should try to use them instead of re-implementing data fetching in useEffect.

JBWatenbergScality (changes requested):
Pure UI review first:
 - the confirm input could be autofocus-ed when the modal is being opened
 - We should add a placeholder mentioning 'type confirm'
 - The status column in the summary should display the error message from the API
 - The table should be sorted by default on the status with the errors firstly displayed
 - For the column VersionId we should have a space between "Version" and "Id"
 - The warning banner icon should be exclamation-circle instead of the triangle (we are deprecating the triangle alternative)
 - The banner about governance bypass should only be displayed for object lock enabled buckets

Base Branch issue: I think this feature should target developement/2.1

General technical review: 
 - We are missing unit tests fro the feature
 - We should try to make the feature redux-free

src/react/databrowser/buckets/details/Overview.tsx
```diff
@@ -103,6 +104,8 @@ function Overview({ bucket, ingestionStates }: Props) {
   const features = useSelector((state: AppState) => state.auth.config.features);
   const { account } = useCurrentAccount();
   const [isErrorModalOpen, setIsErrorModalOpen] = useState(false);
+  const [isEmptyModalOpen, setIsEmptyModalOpen] = useState<boolean>(false);
```
JBWatenbergScality: I think the `Empty Button` it's state and it's accompanying modals should be in the same component. This will ease their testability

src/react/ui-elements/EmptyConfirmation/EmptySummary.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+    return deleteResultData;
+  }, [deleteResult]);
+
+export const EmptySummary = ({
+  deleteResult,
+  onClose,
+  open,
+}: EmptySummaryProps) => {
+  const [isOpen, setIsOpen] = useState<boolean>(open);
```
JBWatenbergScality: The open/onClose props seems to be duplicated with the internal state. I think it can cause a lot of bugs as the props are not synchronised with the internal state. Probably you will want to simply use the props instead and remove the internal state

src/react/ui-elements/EmptyConfirmation/useEmptyBucket.ts
```diff
@@ -0,0 +1,41 @@
...
+            Objects: result,
+          },
+          BypassGovernanceRetention: true,
+        },
+        {
+          onError: (error) => {
+            const err = error as ApiError;
+            setErrorMsg(err?.message);
+          },
+        },
```
JBWatenbergScality: I thinks this can be simplify, we don't need any additional state here. You can simply get the error from the mutation result.
Also getting it from the mutation result will allow you to remove the useCallback here

src/react/queries.ts
```diff
@@ -206,3 +222,60 @@ export const getAccountSeedsQuery = () => ({
...
+          Key: v.Key || '',
+          VersionId: v.VersionId,
+        }),
+      ),
+    ];
+
+    if (deleteObjects?.length) {
+      callback?.(deleteObjects);
+    }
+  },
```
JBWatenbergScality: nit: For future enhancement we might want to move this to a selector. Indeed it will allow to use the same cached data for object listing

Use query have the following props for this purpose:
```
select: (data: TData) => unknown
Optional
This option can be used to transform or select a part of the data returned by the query function. It affects the returned data value, but does not affect what gets stored in the query cache.


```

src/react/ui-elements/EmptyConfirmation/EmptyConfirmationFooter.tsx
```diff
@@ -0,0 +1,46 @@
...
+  approve,
+  cancel,
+  loading = false,
+  isConfirm = false,
+  deleteNumber = 0,
+}: EmptyConfirmationFooterProps) => {
+  return (
+    <Wrap>
+      {loading && (
+        <BasicText>{deleteNumber} objects being deleted...</BasicText>
```
JBWatenbergScality:
```suggestion
        <BasicText>{deleteNumber} objects deleted...</BasicText>
```

src/react/ui-elements/EmptyConfirmation/useEmptyBucket.ts
```diff
@@ -0,0 +1,42 @@
...
+  bucketName: string,
+  callback: (data: S3.DeleteObjectsOutput) => void,
+) => {
+  const { mutateAsync, isLoading, isSuccess, isError, error } =
+    useDeleteObjects();
+
+  const deleteObjectsFn = useCallback(
+    async (result: S3.ObjectIdentifierList) => {
+      try {
+        const mutationResult = await mutateAsync({
```
JBWatenbergScality: why do we use mutateAsync here instead of a classical useMutation ?

src/react/ui-elements/__tests__/EmptyConfirmation.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+  <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>
+);
+jest.mock('../../next-architecture/ui/S3ClientProvider', () => ({
+  useS3Client: jest.fn(),
+}));
+jest.mock('react-redux', () => ({
+  useSelector: () => false,
+}));
+
+describe('EmptyConfirmation', () => {
```
JBWatenbergScality: I think we are missing 3 cases:
- "should enable button once typed confirm"
- "should display count of deleted items while deleting objects"
- "should display the summary when the operation is done"

src/react/ui-elements/__tests__/EmptySummary.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+  });
+
+  it('should render the EmptySummaryList with the correct data', () => {
+    render(
+      <Wrapper>
+        <EmptySummary deleteResult={mockDeleteResult} onClose={() => {}} open />
+      </Wrapper>,
+    );
+
+    expect(screen.getByText('3 objects')).toBeInTheDocument();
```
JBWatenbergScality: We probably need to check more in details the rendered rows

src/react/queries.ts
hervedombya: we both use `select`  and `onSuccess` here because call a callback inside `select` will cause an infinit loop

src/react/ui-elements/DeleteBucket/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const { isBucketEmpty, objectStatus } = useCheckIfBucketEmpty(bucketName);
+  const { mutate: deleteBucket, isLoading } = useDeleteBucket();
+
+  const handleDeleteClick = () => {
+    setIsDeleteModalOpen(!isDeleteModalOpen);
+  };
+
+  const approve = () => {
+    deleteBucket({ Bucket: bucketName });
+    handleDeleteClick();
```
JBWatenbergScality: This should probqbly be done on success of bucket deletion

src/react/ui-elements/EmptyBucket/EmptyBucketSummaryList.tsx
```diff
@@ -0,0 +1,143 @@
...
+const SummaryCell = (
+  value: number,
+  isFailed = false,
+  withIcon = false,
+  messages?: string[],
+) => {
+  const text = maybePluralize(value, 'object');
+  if (withIcon) {
+    return (
+      <Box display="flex" gap={8} alignItems="center">
```
JBWatenbergScality: This can also be achieved with our own stack component from core ui

src/react/ui-elements/EmptyBucket/utils.ts
```diff
@@ -0,0 +1,50 @@
...
+    } else {
+      uniqueMessages[message] = 1;
+    }
+  });
+
+  const uniqueArray = Object.keys(uniqueMessages).map((message) => ({
+    message,
+    errorNumbers: uniqueMessages[message],
+  }));
+
```
JBWatenbergScality: I think this implementation might be simplified thanks to the use of a Set. 

`[...new Set(errorMessages)]` probably does the same

src/react/ui-elements/__tests__/EmptyBucket.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+    await waitFor(() => {
+      expect(selectors.summaryModal()).toBeInTheDocument();
+    });
+  });
+
+  it('button should be disabled when no objects', async () => {
+    jest.spyOn(hooks, 'useCheckIfBucketEmpty').mockReturnValue({
+      isBucketEmpty: false,
+      objectStatus: 'success',
+    });
```
JBWatenbergScality: I would instead prefer to re-mock using server.use via MSW

src/react/ui-elements/__tests__/EmptyBucket.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+    userEvent.click(selectors.confirmButton());
+
+    await waitFor(() => {
+      expect(selectors.deletionLoading()).toBeInTheDocument();
+    });
+  });
+
+  it('fetch the data when user approve', async () => {
+    reduxRender(<EmptyBucket bucketName={bucketName} />);
```
JBWatenbergScality, line 113: I think we are missing some tests case in case the deletion fail

src/react/ui-elements/__tests__/EmptyBucket.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+    });
+
+    await waitForElementToBeRemoved(
+      () => screen.getByText(/Deletion in progress/i),
+      { timeout: 10_000 },
+    );
+
+    await waitFor(() => {
+      expect(selectors.summaryModal()).toBeInTheDocument();
+    });
```
JBWatenbergScality, line 178: We should also check that we have the right numbers

src/react/ui-elements/DeleteBucket/index.tsx
hervedombya: Like this : ` deleteBucket({ Bucket: bucketName }, {
      onSuccess: () =>  {
        handleDeleteClick()
      }
    });` ?

src/react/ui-elements/EmptyBucket/utils.ts
hervedombya: I don't want to juste remove duplicates but create an array with unique message and his occurrence. Ex : if I have this array : `['error', 'error', 'error', 'error', 'error']`, I want to create an array of objects like this : `[{ message: 'error', errorNumbers: 5}]`

Actually I can maybe use array.reduce instead of forEach and Object.keys

Cuervino:
1. I think we use the character " instead of ' in the UI content, so the modal title could be adapted.

2. At the Summary step, the button should be in Primary style, not Outline.

3. I would change the notification banner text like that:
> The action of emptying can erase a maximum of 20 000 objects.
> For buckets with a substantial number of objects, an expiration workflow (lifecycle rule) could provide an effective alternative for emptying the bucket.

Cuervino (approved):
Few minor comments.

hervedombya:
> 1. I think we use the character " instead of ' in the UI content, so the modal title could be adapted.
> 2. At the Summary step, the button should be in Primary style, not Outline.
> 3. I would change the notification banner text like that:
> 
> > The action of emptying can erase a maximum of 20 000 objects.
> > For buckets with a substantial number of objects, an expiration workflow (lifecycle rule) could provide an effective alternative for emptying the bucket.

JBWatenbergScality (approved):
LGTM !!

PR #611, opened 2023-06-15
Fix several issues for Non Storage Manager Users, allow sorting Data Used Columns, fix cold storage and object restoration issues

JBWatenbergScality:


ChengYanJin (approved):
LGTM!

PR #625, opened 2023-07-25
ZKUI-372 // disable empty button after delete

hervedombya:


JBWatenbergScality (commented):
Can you cover this with a test ?

JBWatenbergScality (approved):

PR #630, opened 2023-08-23
ZKUI-374 // The list versions toggle inactive without page refresh

hervedombya:


src/react/databrowser/buckets/details/__tests__/Overview.test.tsx
```diff
@@ -212,3 +212,56 @@ describe('Overview', () => {
...
+import { rest } from 'msw';
+import { setupServer } from 'msw/node';
+import { BUCKET_NAME } from '../../../../actions/__tests__/utils/testUtil';
+import { TEST_API_BASE_URL } from '../../../../utils/testUtil';
+const mockResponse =
+  '<VersioningConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><Status>Enabled</Status></VersioningConfiguration>';
+
+const server = setupServer();
+beforeAll(() => {
+  server.listen({ onUnhandledRequest: 'error' });
```
ChengYanJin, line 278: We have some fetch error from MSW when running the Overview Test which bring a lot of noisy. I would suggest to mock the necessary call to resolve the problem.

```js    
FetchError {
      message: 'request to http://testendpoint/ failed, reason: [MSW] Cannot bypass a request when using the "error" strategy for the "onUnhandledRequest" option.',
      type: 'system',
      errno: undefined,
      code: undefined
    }
```

ChengYanJin (commented):
Also since it's bug fix, I would suggest we target development/2.0 branch.

PR #634, opened 2023-09-04
ZKUI-376: fix the workflow not visible on workflow page

ChengYanJin:
The value of the form was not set correctly which cause the issue.

So we need to do reset the form value once we load workflow.


JBWatenbergScality (approved):
LGTM !

PR #641, opened 2023-09-15
ui: remove xcore location as the destination of the transition workflow

ChengYanJin:


ChengYanJin:
will reopen a PR target developement/2.0

PR #644, opened 2023-09-20
Fix the page fully reload upon token renewal and assume role after account creation

ChengYanJin:
This PR fixed two issues explained below:

1. Zenko UI gets fully reload upon token renewal is caused by the wrongly added `token` within queryKey when fetching `WebIdentityRoles`. However it is totally unnecessary.

2. After account creation, we switch to the new account so we should assume role again which is missing before, cause the issue that we still load the IAM User/ policy of previous account.


JBWatenbergScality (approved):
LGTM !

PR #664, opened 2023-10-26
ZKUI-387: No basepath for development when initing s3 and iam client

ChengYanJin:
In order to run all the UIs in local development


JBWatenbergScality (approved):
LGTM !!!

PR #667, opened 2023-11-06
ZKUI-397: refactor bucket overview for Veeam use case

ChengYanJin:
Veeam Bucket: 
image

Not Veeam Bucket:
image

Error when loading bucket tagging failed 
image

### DIFF
--- a/src/js/mock/S3ClientMSWHandlers.ts
+++ b/src/js/mock/S3ClientMSWHandlers.ts
@@ -118,6 +118,43 @@ export function mockBucketOperations(
         );
       }
 
+      if (req.url.searchParams.has('cors')) {
+        return res(
+          ctx.xml(`
+          <?xml version="1.0" encoding="UTF-8"?>
+          <CORSConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
+          </CORSConfiguration>
+        `),
+        );
+      }
+
+      if (req.url.searchParams.has('acl')) {
+        return res(
+          ctx.xml(`
+          <?xml version="1.0" encoding="UTF-8"?>
+          <AccessControlPolicy xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
+            <Owner>
+              <ID>1234</ID>
+              <DisplayName>test</DisplayName>
+            </Owner>
+            <AccessControlList>
+            </AccessControlList>
+          </AccessControlPolicy>
+        `),
+        );
+      }
+
+      if (req.url.searchParams.has('object-lock')) {
+        return res(
+          ctx.xml(`
+          <?xml version="1.0" encoding="UTF-8"?>
+          <ObjectLockConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
+            <ObjectLockEnabled>Enabled</ObjectLockEnabled>
+          </ObjectLockConfiguration>
+        `),
+        );
+      }
+
       return res(ctx.status(404));
     },
   );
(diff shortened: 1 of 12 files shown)


src/js/mock/S3ClientMSWHandlers.ts
```diff
@@ -178,3 +178,33 @@ export const mockObjectEmpty = (bucketName: string) => {
...
+  return rest.get(
+    `${zenkoUITestConfig.zenkoEndpoint}/${bucketName}`,
+    (req, res, ctx) => {
+      if (req.url.searchParams.has('tagging')) {
+        return res(
+          ctx.xml(`
+          <?xml version="1.0" encoding="UTF-8"?>
+          <Tagging>
+            <TagSet>
+              <Tag><Key>Usecase</Key><Value>Veeam 12</Value></Tag>
```
JBWatenbergScality: Shall we prefix the tag Key with something like `X-Scality` to warn the user that this is an internal tag, and later on disable the edition of the X-Scality tag in the "bucket tag" ui ?

src/react/databrowser/buckets/details/Overview.tsx
```diff
@@ -89,6 +96,15 @@ function Overview({ bucket, ingestionStates }: Props) {
   const { account } = useCurrentAccount();
   const [isErrorModalOpen, setIsErrorModalOpen] = useState(false);
 
+  const { tags } = useBucketTagging({ bucketName: bucket.name });
```
JBWatenbergScality, line 106: What happens in case of failure to retrieve the tags ? For now IIUC we simply ignore the issue. Do we want to handle this ?

src/react/next-architecture/domain/business/buckets.ts
```diff
@@ -91,6 +92,15 @@ export const queries = {
...
   }),
+  getBucketTagging: (s3Client: S3, bucketName?: string) => ({
+    queryKey: ['bucketTagging', getS3ClientHash(s3Client), bucketName],
+    queryFn: () =>
+      s3Client
+        .getBucketTagging({ Bucket: notFalsyTypeGuard(bucketName) })
+        .promise(),
+    enabled: !!bucketName && !!s3Client.config.credentials?.accessKeyId,
+    ...noRefetchOptions,
+  }),
```
JBWatenbergScality: I wonder why the bucketName parameter is optional ? Wouldn't it simplify everything to make it mandatory ? Allowing to drop the notFalsyTypeGuard and the check in the enabled

src/js/mock/S3ClientMSWHandlers.ts
ChengYanJin: Good point !

src/react/utils/testUtil.tsx
```diff
@@ -144,6 +144,14 @@ export const queryClient = new QueryClient({
   },
 });
 
+//Note: React Query version 4 setLogger function is removed.
+setLogger({
+  log: console.log,
+  warn: console.warn,
+  // ✅ no more errors on the console
+  error: () => {},
```
JBWatenbergScality, line 155: nit: I think we need to add an eslint ignore comment here

src/react/databrowser/buckets/details/Overview.tsx
```diff
@@ -136,6 +161,13 @@ function Overview({ bucket, ingestionStates }: Props) {
             : null
         }
       />
+      <Toast
+        message="Encountered issues loading bucket tagging, causing uncertainty about the use-case. Please refresh the page."
+        open={tags.status === 'error' && bucketTaggingToast}
+        onClose={() => {
+          setBucketTaggingToast(false);
```
JBWatenbergScality, line 169: I think we can 'reset' the query instead of introducing a new state

src/react/databrowser/buckets/details/__tests__/Overview.test.tsx
```diff
@@ -235,13 +241,29 @@ const server = setupServer(
...
 
+const selectors = {
+  editDefaultRetentionButton: () =>
+    screen.getByRole('button', {
+      name: /edit default retention/i,
+    }),
+  bucketTaggingErrorToastCloseButton: () =>
+    screen.getByRole('button', {
+      name: /close/i,
+    }),
```
JBWatenbergScality: Nit: I would add a "within" check to be sure we select the right close button

src/react/databrowser/buckets/details/__tests__/Overview.test.tsx
```diff
@@ -235,13 +241,29 @@ const server = setupServer(
...
+  bucketTaggingErrorToast: () =>
+    screen.getByText(
```
JBWatenbergScality: I would ideally select by role and name

src/react/databrowser/buckets/details/__tests__/Overview.test.tsx
```diff
@@ -273,4 +294,43 @@ describe('Overview', () => {
       expect(useUpdateBucketVersioningMock).toHaveBeenCalledWith(mockResponse);
     });
   });
+
+  it('should display the Veeam use-case and disable the edition of default retention', async () => {
+    //Setup
+    server.use(mockGetBucketTagging(bucketName));
+    //Exersise
```
JBWatenbergScality: nit: typo
```suggestion
    //Exercise
```

JBWatenbergScality (approved):
LGTM !

PR #678, opened 2023-11-16
ZKUI-390: Prepare all the mutation needed for Veeam

ChengYanJin:


src/js/mutations.ts
```diff
@@ -0,0 +1,143 @@
...
+
+// const roleArn = `arn:aws:iam::${accountId}:role/storage-manager-role`;
+const useAssumeRoleMutation = () => {
+  const { getQuery } = useAssumeRoleQuery();
+  return useMutation({
+    mutationFn: (roleArn: string) => getQuery(roleArn).queryFn(),
+    onSuccess(_, variables) {
+      setRoleArnStored(variables);
+    },
+  });
```
JBWatenbergScality: I would useSetAssumeRole instead and return setRole among with it's result, also I would make setRole return the mutation object (without the mutate method, which would be replaced by the setRole one).

JBWatenbergScality (approved):
LGTM !!!

PR #711, opened 2023-12-15
ZKUI-426 // Polishing ui

hervedombya:


src/react/ui-elements/Veeam/VeeamSkipModal.tsx
```diff
@@ -0,0 +1,54 @@
...
+  exitAction: () => void;
+  modalContent: JSX.Element;
+  title?: string;
+};
+
+const ModalContent = styled.div`
+  max-width: 30rem;
+`;
+
+export const VeeamSkipModal = ({
```
JBWatenbergScality, line 17: nit: Maybe we should add a test to make sure this confirm modal is opened when clicking on `Skip Use case configuration `

hervedombya: Done in f00d1b3fc0b1d110e9937196d6dd4f490010f362

JBWatenbergScality (approved):
LGTM !

PR #713, opened 2023-12-19
Improvement/zk UI 430- removing defaultTheme and brand

JeanMarcMilletScality:


src/types/styled.ts
```diff
@@ -1,80 +1,7 @@
...
-      backgroundLevel4: string;
-      textPrimary: string;
-      textSecondary: string;
-      textTertiary: string;
-      textReverse: string;
-      textLink: string;
-    };
-  }
+  // eslint-disable-next-line @typescript-eslint/no-empty-interface
+  export interface DefaultTheme extends CoreUITheme {}
```
JBWatenbergScality: Do we really need this one ? Can't we inherit from the one defined in core ui ?

JBWatenbergScality (approved):
LGTM !

PR #716, opened 2024-01-10
ARTESCA-10663: replication should not be possible from source veeam tagged bucket

ChengYanJin:


src/react/workflow/SourceBucketOption.tsx
```diff
@@ -107,13 +110,17 @@ export const SourceBucketOption = ({
...
+  const { tags } = useBucketTagging({ bucketName });
 
   const isOptionDisabled = disableOption
     ? disableOption({
         isBucketVersionned:
           versionning.status === 'success' && versionning.value === 'Enabled',
         locationInfos:
           locationInfos.status === 'success' ? locationInfos.value : undefined,
+        isVeeamBucket:
+          tags.status === 'success' &&
```
JBWatenbergScality: Shouldn't we also disable while it is in "loading" status ?

JBWatenbergScality: If it is in error shall we toast ?

JBWatenbergScality (approved):

PR #721, opened 2024-01-18
ARTESCA-9464 // unexpected replication workflows

hervedombya:


src/react/workflow/DeleteWorkflowButton.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+const getWorkflowType = (workflow: WorkflowType) => {
+  return isTransitionWorkflow(workflow)
+    ? WorkflowRule.Transition
+    : isExpirationWorkflow(workflow)
+    ? WorkflowRule.Expiration
+    : WorkflowRule.Replication;
+};
+
+export function useFilteredReplicationWorkflows(
```
JBWatenbergScality: nit: we have a generic naming approach here where in fact the function "only" filters the workflow having the same source bucket name. I vote for either renaming it accordingly to staet that it is filtering only workflows having the same source bucket, either (and maybe a better approach) use a reverse control approach and refactor this hook so that it takes a predicate function as an argument `(replication: Replication) => boolean`

src/react/workflow/DeleteWorkflowButton.tsx
```diff
@@ -0,0 +1,189 @@
...
+    () =>
+      workflows.length
+        ? workflows.map((wf) => getWorflowName(wf))
+        : [getWorflowName(workflow)],
+    [workflows],
+  );
+
+  const handleDeleteWorkflow = useCallback(() => {
+    mutate(workflow);
+    setOpenDeleteModal(false);
```
JBWatenbergScality: We may want to close the modal only once the deletion is completed and change the button to loading state while the deletion is ongoing. Also display a toast if the deletion failed and keep the modal opened with the button back to enabled so that the user can easily retry.

JBWatenbergScality (approved):
LGTM !

PR #729, opened 2024-02-07
ARTESCA-11299: prevent create a new storage service and edition of the default

ChengYanJin:


src/react/locations/LocationEditor.tsx
```diff
@@ -73,9 +74,14 @@ function LocationEditor() {
     convertToForm({ ...newLocationDetails(), ...locationEditing }),
   );
   const selectOptions = useMemo(() => {
-    //@ts-expect-error fix this when you are working on it
-    return selectStorageOptions(capabilities, makeLabel, !editingExisting);
-  }, [capabilities, editingExisting]);
+    if (!locations) return [];
```
JBWatenbergScality: Are we sure we don't want to display anything in this list if no locations exists yet ? In theory we shouldn't end up in this case except some specific install, however i think we should display all the locations options if none are created.

JBWatenbergScality (approved):
LGTM !

PR #731, opened 2024-02-20
ARTESCA-11467: fix issue with xcore dependency for veeam assistant

hervedombya:


src/react/ui-elements/Veeam/VeeamConfiguration.tsx
```diff
@@ -146,9 +149,11 @@ const Configuration = () => {
   // clear server errors if clicked on outside of element.
   const formRef = useRef(null);
   const xCoreConfig = useXcoreConfig('run');
-  const { useClusterCapacity } = useXCoreLibrary();
+  const xCoreLibrary = useXCoreLibrary();
+  const { useClusterCapacity } =
+    xCoreLibrary === XCORE_NOT_AVAILABLE ? () => ({}) : xCoreLibrary;
   const { clusterCapacity, clusterCapacityStatus } =
-    useClusterCapacity(xCoreConfig);
+    useClusterCapacity?.(xCoreConfig) || {};
```
JBWatenbergScality: This one is breaking the rule of hooks, we cannot have conditional hooks 😢 . We will need to introduce a new component doing this based on the condition

JBWatenbergScality (changes requested):

JBWatenbergScality (approved):
LGTM !!

PR #732, opened 2024-02-21
ARTESCA-11338: Fix the issue of Veeam welcome modal

ChengYanJin:
We change the condition to display the Veeam Welcome Modal from No Veeam Account to No Account at all.

### DIFF
--- a/src/react/ui-elements/Veeam/VeeamWelcomeModal.tsx
+++ b/src/react/ui-elements/Veeam/VeeamWelcomeModal.tsx
@@ -12,7 +12,6 @@ import {
 import { ArtescaLogo } from './ArtescaLogo';
 import { VeeamLogo } from './VeeamLogo';
 import { useAccounts, useAuthGroups } from '../../utils/hooks';
-import { VEEAM_DEFAULT_ACCOUNT_NAME } from './VeeamConstants';
 import {
   getSkipVeeamAssistant,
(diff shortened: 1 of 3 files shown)


src/react/ui-elements/Veeam/VeeamWelcomeModal.tsx
```diff
@@ -25,37 +24,51 @@ const CustomModal = styled(Modal)`
...
    3. Storage Manager is logged in
    4. Have never click on the skip button
+   5. Not already in the Veeam configuration view
    */
   const isVeeamWelcomeModalEnabled =
     features.includes(VEEAM_FEATURE) &&
     isStorageManager &&
-    !isVeemAccountCreated &&
-    !isVeeamAssistantSkipped;
+    isZeroAccountCreated &&
```
JBWatenbergScality, line 46: I feel like when we are in " loading status" for the account there is a possibility to have the modal displayed, and then once the loading is finished the modal is dismounted

ChengYanJin: I believe the isZeroAccountCreated is false while loading the account? 

 ```js
 const isZeroAccountCreated = status === 'success' && accounts.length === 0;
 ```

JBWatenbergScality (approved):
LGTM !

PR #733, opened 2024-02-23
Improvement/artesca 11213 bucket creation tooltip

JeanMarcMilletScality:


src/react/workflow/ReplicationForm.tsx
```diff
@@ -286,14 +286,34 @@ function ReplicationForm({ prefix = '', isCreateMode, ...props }: Props) {
...
-                            ?.supportsReplicationSource ||
-                          !isBucketVersionned ||
-                          isVeeamBucket
-                        }
+                        }) => ({
+                          disabled:
+                            isReplicationSourceSupported ||
+                            !isBucketVersionned ||
+                            isVeeamBucket,
+                          disabledMessage: isReplicationSourceSupported ? (
```
JBWatenbergScality, line 296: Can we extend the existing workflow tests to check the presence of the tooltip ?

src/react/workflow/SourceBucketOption.tsx
```diff
@@ -17,9 +16,9 @@ import { useEffect } from 'react';
 type DisableOptionProps = {
   disableOption?: (obj: {
     isBucketVersionned?: boolean;
-    locationInfos?: LocationStorageInfos;
+    isReplicationSourceSupported?: boolean;
     isVeeamBucket: boolean;
-  }) => boolean;
+  }) => { disabled: boolean; disabledMessage?: JSX.Element };
```
hervedombya, line 21: type `ReactNode` is better

JeanMarcMilletScality: We only want a string or an ul element. Maybe we should use a union type ?

JBWatenbergScality (approved):
LGTM !

PR #738, opened 2024-03-18
ARTESCA-11642: Skip until next login

hervedombya:


src/react/ui-elements/Veeam/VeeamWelcomeModal.tsx
```diff
@@ -45,7 +47,8 @@ export const VeeamWelcomeModalInternal = () => {
     isStorageManager &&
     isZeroAccountCreated &&
     !isVeeamAssistantSkipped &&
```
ChengYanJin: do we still need `isVeeamAssistantSkipped` ? I think we dont need this state in the localStorage anymore.

We should review the comments starting from line 38 since we change the logic.

src/react/ui-elements/Veeam/useNextLogin.test.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    mockGetSessionState.mockReturnValue(sessionState);
+
+    const { result } = renderHook(() => useNextLogin(), {
+      wrapper: NewWrapper(),
+    });
+
+    expect(result.current.isNextLogin).toBe(false);
+  });
+});
```
ChengYanJin, line 46: there is another scenario which is "it should return true when where is no session_state in local storage"

src/react/ui-elements/Veeam/VeeamWelcomeModal.tsx
```diff
@@ -101,9 +106,7 @@ const VeeamModalComponent = () => {
               label={'Skip'}
```
ChengYanJin:
```suggestion
              label={'Skip until the next login'}
```

ChengYanJin (approved):
LGTM!

PR #757, opened 2024-06-27
bugifx: Update SelectAccountIAMRole component

hervedombya:


src/react/ui-elements/SelectAccountIAMRole.tsx
```diff
@@ -289,10 +291,14 @@ export const _SelectAccountIAMRole = (props: SelectAccountIAMRoleProps) => {
...
 };
 
-export const SelectAccountIAMRole = (props: SelectAccountIAMRoleProps) => {
+export default function SelectAccountIAMRole(props: SelectAccountIAMRoleProps) {
   return (
-    <InternalProvider defaultValue={props.defaultValue}>
-      <_SelectAccountIAMRole {...props} />
-    </InternalProvider>
+    <ConfigProvider>
+      <AuthProvider>
```
MonPote, line 307: You should put these import in the `InternalProvider`  component

MonPote (changes requested):

src/react/ui-elements/SelectAccountIAMRole.tsx
hervedombya: Fixed in 8ded7cb86bd1a25d575d57b3b1b96f5032f5155d
Bring this providers seems to break tests and the implementation of the component

MonPote (approved):

PR #788, opened 2024-10-23
fix issue with Tooltip being cut

JeanMarcMilletScality:
Fix cold storage being cut in object list (add textwrap and align left)
Bonus : 
- Add margin after folder icon
- improve display of storage cell)

Modified on
image


JBWatenbergScality (approved):
LGTM !

PR #791, opened 2024-12-16
Bugfix/artesca 14165 capacity decimals issue

JeanMarcMilletScality:
Issue with JS floating number precision

### DIFF
--- a/src/react/ui-elements/Veeam/VeeamCapacityModal.test.tsx
+++ b/src/react/ui-elements/Veeam/VeeamCapacityModal.test.tsx
@@ -5,6 +5,7 @@ import { bucketName } from '../../../js/mock/S3Client';
 import { NewWrapper, TEST_API_BASE_URL } from '../../utils/testUtil';
 import { VeeamCapacityModal } from './VeeamCapacityModal';
 import { VEEAM_XML_PREFIX } from './VeeamConstants';
+import userEvent from '@testing-library/user-event';
 
 describe('VeeamCapacityModal', () => {
   const mockMutate = jest.fn();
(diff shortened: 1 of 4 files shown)


src/react/ui-elements/Veeam/VeeamCapacityModal.tsx
```diff
@@ -29,10 +29,14 @@ const schema = Joi.object({
     .min(1)
     .max(1024)
     .custom((value, helpers) => {
-      if (!Number.isInteger(value * 100)) {
-        return helpers.message({
-          custom: '"capacity" must have at most 2 decimals',
-        });
+      const stringValue = value.toString();
+      if (stringValue.includes('.')) {
```
MonPote: I think you should add a test for that

src/react/ui-elements/Veeam/VeeamCapacityModal.test.tsx
```diff
@@ -52,6 +53,16 @@ describe('VeeamCapacityModal', () => {
...
 
+  it('should enabled edit button when value is valid', async () => {
+    await userEvent.click(selectors.editBtn());
+
+    await userEvent.clear(selectors.capacityInput());
+    expect(selectors.editModalBtn()).toBeDisabled();
+    await userEvent.type(selectors.capacityInput(), '2.2');
+
+    expect(selectors.editModalBtn()).toBeEnabled();
+  });
```
JBWatenbergScality, line 64: nit: aren't we missing a test to check that numbers with more than 2 decimals are invalid ?

JBWatenbergScality (approved):

src/react/ui-elements/Veeam/VeeamCapacityModal.test.tsx
JeanMarcMilletScality: There is a test for this:
`  it('should validate capacity value correctly : number with more than 2 decimals', async () => `

PR #801, opened 2025-03-10
Bugfix/artesca 14779 veeam app order

JeanMarcMilletScality:
Change veeam app order in isv connector modal
Fix cache issue on Policy creation


JBWatenbergScality (approved):
lgtm

PR #826, opened 2025-04-03
Add banner notification in LocationEditor for storage location limits

ZiyangLinScality:


JBWatenbergScality (commented):
I would add a unit test for this case

JBWatenbergScality (approved):

PR #833, opened 2025-04-10
Fix hidden Immutable option

ZiyangLinScality:


src/react/ISV/components/ISVSummary.tsx
```diff
@@ -78,6 +79,8 @@ export const ISVSummary = ({
     application: application,
   });
 
+  const shouldHideImmutableSection = application === VEEAM_OFFICE_365;
```
JBWatenbergScality, line 82: Only for V6 and 7 of this app not V8+

JBWatenbergScality (changes requested):

JBWatenbergScality (approved):

PR #862, opened 2025-04-30
ZKUI-437 // Disable/Warn lifecycle expiration workflow for Veeam buckets

hervedombya:


src/react/workflow/ExpirationForm.tsx
```diff
@@ -198,6 +224,33 @@ export function ExpirationForm({ prefix = '' }: Props) {
...
             }
           />
+
+          {isISVBucket && (
+            <Stack direction="vertical" gap="r16">
+              <Banner
+                icon={<Icon name="Exclamation-circle" color="statusWarning" />}
+                variant="warning"
+              >
+                This bucket is tagged as being used in a ISV use-case.
```
JBWatenbergScality: @Cuervino I think ISV is not the public word we use, better use Connector here no, wdyt ?
```suggestion
                This bucket is tagged as being used in a connector use-case.
```

src/react/workflow/ExpirationForm.tsx
```diff
@@ -144,6 +155,21 @@ export function ExpirationForm({ prefix = '' }: Props) {
...
 
+  const { tags } = useBucketTagging({ bucketName: sourceBucketName });
+  const isISVBucket =
+    tags.status === 'success' &&
+    (tags.value?.[BUCKET_TAG_APPLICATION] ||
+      tags.value?.[BUCKET_TAG_VEEAM_APPLICATION]);
+
+  useMemo(() => {
+    if (isISVBucket) {
+      setValue(`${prefix}understandISVRisk`, false);
```
JBWatenbergScality: I think there is an option for revalidating instead of calling `trigger()`

src/react/workflow/__tests__/ExpirationForm.test.tsx
```diff
@@ -216,4 +217,57 @@ describe('ExpirationForm', () => {
     const formValidation = screen.getByTestId('form-expiration');
     expect(formValidation.textContent).toBe('form-valid');
   });
+
+  it('should validate ISV bucket confirmation checkbox', async () => {
```
JBWatenbergScality, line 221: We should add another test to check that the checkbox is not there and the form can be submitted for non isv buckets

JBWatenbergScality (approved):

PR #892, opened 2025-06-18
ARTESCA-15484: Add logic to disable endpoint only if there is only mo…

ChengYanJin:
…re addition veeam s3 endpoint

### DIFF
--- a/src/react/endpoint/EndpointList.tsx
+++ b/src/react/endpoint/EndpointList.tsx
@@ -1,4 +1,11 @@
-import { ConstrainedText, Icon, Wrap, spacing } from '@scality/core-ui';
+import {
+  ConstrainedText,
+  Icon,
+  Loader,
+  Text,
+  Wrap,
+  spacing,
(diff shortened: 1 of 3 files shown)


src/react/endpoint/EndpointList.tsx
```diff
@@ -109,18 +122,26 @@ function EndpointList({ endpoints, locations }: Props) {
...
-            <DeleteEndpoint
-              hostname={original.hostname}
-              isBuiltin={original.isBuiltin}
-            />
-          );
+          if (endpointsDeletionDisabledStatus === 'success') {
+            return (
+              <DeleteEndpoint
+                hostname={original.hostname}
+                endpointsDeletionDisabledMap={endpointsDeletionDisabledMap}
```
JBWatenbergScality:
```suggestion
                disabled={endpointsDeletionDisabledMap[original.hostname]}
```

JBWatenbergScality (approved):

PR #904, opened 2025-06-30
RING-50810 // create script for manage proxy config

hervedombya:


deploy-script.sh
```diff
@@ -0,0 +1,104 @@
...
+fi
+
+echo "Generating nginx configuration..."
+
+# Read existing nginx config and prepare for modification
+cp "$NGINX_CONFIG" "$TEMP_CONFIG"
+
+# Find the position to insert proxy locations (before the closing brace)
+# Remove the last closing brace temporarily
+sed -i '$d' "$TEMP_CONFIG"
```
JBWatenbergScality: In theory nothing guarantees that the Server block in nginx config is the last one. Can we make this script "detect" the Server block and insert the generated rules in there ?

.github/scripts/test-deploy-script.sh
```diff
@@ -0,0 +1,222 @@
...
+      while ((getline proxy_line < "'$PROXY_TEMP'") > 0) {
+        print proxy_line
+      }
+      close("'$PROXY_TEMP'")
+      in_server = 0
+    }
+  }
+  
+  { print }
+  ' "$TEMP_CONFIG" > "${TEMP_CONFIG}.new"
```
JBWatenbergScality: Not sure why do we have this re-implemented in the test ?

JBWatenbergScality (approved):

PR #988, opened 2025-11-05
ARTESCA-16316 - remove certificate from truststore

ChengYanJin:
Screenshot 2025-11-05 at 17 45 07


src/react/ui-elements/DeleteConfirmation.tsx
```diff
@@ -42,7 +42,7 @@ const DeleteConfirmation = ({
               variant="danger"
               onClick={() => approve()}
               icon={(isLoading || loading) && <Loader size="larger" />}
```
JeanMarcMilletScality: There is a isLoading prop for Button that disable the button + change icon to spinner directly :wink:

JeanMarcMilletScality (approved):

PR #990, opened 2025-11-06
Download certificate in modal

JeanMarcMilletScality:
Add download functionality for certificates in CertificateDetails component

- Introduced a download button for each certificate, allowing users to download the certificate in PEM format.
- Updated the CertificateDetails component to include the new button and associated download logic.
- Modified hooks and types to support the new functionality, ensuring the original PEM is accessible.
- Adjusted tests to verify the presence of the download button in the CertificateDetails modal.

image


ChengYanJin (approved):
LGTM!

PR #1001, opened 2025-12-03
Replace data browser by using data-browser-library

ZiyangLinScality:
This is the first step to integrate, just for the UI component part.

The next step will replace the existing S3 API requests with the library’s hook.

At the same time fixed 2 issues:
1. [Setting default alertManagerUrl in AlertProvider](https://github.com/scality/zenko-ui/commit/7cc2c0bb746b13b747d44abd74c6df2d212b511d);
2. [Moving ISVStep Context to a separate file resolves the nested reference issue.](https://github.com/scality/zenko-ui/commit/3993dbba093b850da17ac8c04c5980136dea069d)

### DIFF
--- a/src/react/databrowser/DataBrowser.tsx
+++ b/src/react/databrowser/DataBrowser.tsx
@@ -19,6 +19,7 @@ import { LocationSection } from './buckets/LocationSection';
 import { StartISVConnectorButton } from '../ISV/components/StartISVConnectorButton';
 import { useDataBrowserS3Config } from './hooks/useDataBrowserS3Config';
 import { BucketMetricsPrefetch } from './hooks/useBucketMetrics';
+import { BucketLocationsPrefetch } from './BucketLocationsPrefetch';
 
 const EXTRA_BUCKET_OVERVIEW_SECTIONS = [
   {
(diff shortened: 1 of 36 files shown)


src/react/databrowser/DataBrowser.tsx
```diff
@@ -1,113 +1,122 @@
...
+      </Box>
+    ),
+    [pathname, prefixPath, accountName, basePath],
+  );
+
+  // TODO: Move DataBrowserProvider to the S3ClientProvider when refactoring the S3ClientProvider
+  return (
+    <DataBrowserProvider getS3Config={getS3Config} theme={theme}>
+      <BucketMetricsPrefetch />
+      <BucketLocationsPrefetch />
```
JBWatenbergScality, line 111: It's not super clear in my mind why do we need this.

ZiyangLinScality: These two can improve loading efficiency because our metrics interface can accept multiple bucket information transfers, and the location part can be pre-loaded so that there won't be any loading UI displayed when the user scrolls.

JBWatenbergScality (approved):

PR #1002, opened 2025-12-04
Add bucket name and role fields

JBWatenbergScality:


JBWatenbergScality:
Wrong change need to be done on replication workflow

PR #1003, opened 2025-12-05
Replace ISV S3 operations with @scality/data-browser-library (AWS SDK v3)

ZiyangLinScality:
Added three new providers to support the library integration:
- `IAMProvider` - Provides IAM client context
- `ZenkoProvider` - Provides Zenko client context   (still need zenokclient because the location)
- `DataBrowserProvider` - Wraps data-browser-library provider

Migrated ISV modules to use library hooks:
- Bucket creation with Object Lock → `useCreateBucket()`
- Bucket tagging → `useSetBucketTagging()`
- Object upload (Veeam XMLs) → `usePutObject()`

Ignore some tests, will fix them when cleanup the old databrowser and workflow


src/js/mutations.ts
```diff
@@ -312,47 +325,119 @@ const useAttachPolicyToUserMutation = () => {
   });
 };
 
-const usePutBucketTaggingMutation = () => {
-  const s3Client = useS3Client();
-  return useMutation(
-    ({ bucketName, tagSet }: { bucketName: string; tagSet: TagSetItem[] }) => {
-      return s3Client
-        .putBucketTagging({
+const usePutBucketTaggingMutationWithConfig = () => {
```
JBWatenbergScality: Shouldn't we do this through exposed library from the data browser to correctly handle cache invalidation and other internals of Data Browser library ?

src/js/mutations.ts
```diff
@@ -312,47 +325,119 @@ const useAttachPolicyToUserMutation = () => {
...
+      s3Config: S3Config;
+      Bucket: string;
+      Key: string;
+      Body: string;
+      ContentType?: string;
+    }) => {
+      const client = createCustomS3Client(s3Config);
+
+      return client.send(
+        new PutObjectCommand({
```
JBWatenbergScality: Same as above here

src/react/ISV/modules/veeam/components/VeeamCapacityOverviewRow.tsx
```diff
@@ -13,32 +8,86 @@ import {
...
+          setXml('');
+        }
+      } catch (error) {
+        console.error('[VeeamCapacityOverviewRow] Error parsing Body:', error);
+        setXml('');
+      }
+    };
+
+    parseBody();
+  }, [veeamObject]);
```
JBWatenbergScality: This shouldn't be an extra state and effect as it can be directly derived from another state, it can probably simply be migrated to a useMemo.
It looks like we are trying many different format because this code is likely generated from type definition of the SDK but in practice it will always output the same format deterministically, so it can probably be simplified to only what you really get

src/react/ISV/utils/s3Operations.ts
```diff
@@ -0,0 +1,39 @@
...
+  return client.send(new PutBucketTaggingCommand(params));
+};
+
+export const putObjectWithConfig = async (
+  config: S3Config,
+  params: PutObjectCommandInput,
+): Promise<PutObjectCommandOutput> => {
+  const client = createS3Client(config);
+  return client.send(new PutObjectCommand(params));
+};
```
JBWatenbergScality: Same issue as above I think this should use exposed utilities from data browser because else it gets dis-connected with data browser internals and we get cache compatibilities issues or state mismatches

src/react/utils/index.ts
ZiyangLinScality, line 1: These codes are copied from S3Client.ts(will be removed)

JBWatenbergScality (approved):

PR #1007, opened 2025-12-10
add Veeam auto-repository foundation types and feature flag

hervedombya:
- `ISVConfig` extended with:
`autoCreateRepository?: boolean` - Toggle for automatic repository creation
`immutablePeriodDays?: number `- Retention period for immutable backups

- `VeeamRepositoryData` - Matches the summary needed data

Screenshot 2025-12-15 at 15 56 14

### DIFF
--- a/src/react/ISV/types/index.ts
+++ b/src/react/ISV/types/index.ts
@@ -42,6 +42,8 @@ export type ISVConfig = {
   application?: string;
   enableImmutableBackup?: boolean;
   buckets?: Bucket[];
+  autoCreateRepository?: boolean;
+  immutablePeriodDays?: number;
 };
 
 export type ISVFieldOverride = {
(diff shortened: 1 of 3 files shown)


src/react/ISV/types/index.ts
```diff
@@ -92,6 +94,16 @@ export type ISVSummaryData = {
   accessKey: string;
   secretKey: string;
   application: string;
+  repositoryData?: VeeamRepositoryData;
+};
+
+export type VeeamRepositoryData = {
+  repositoryId?: string;
+  repositoryName?: string;
+  status: string;
```
ZiyangLinScality: I think status can use union type will be more type-safe, but I saw backend side you use string

ZiyangLinScality (approved):
LGTM

PR #1011, opened 2025-12-15
implement capacity calculations and parsing functions with unit tests

hervedombya:
When creating a Veeam repository, the API requires:
- `storageConsumptionLimitKind`: `'TB'` or `'PB'`
- `storageConsumptionLimitCount`: integer count

The form provides capacity as bytes in string. These utilities:
- Parse string bytes to integer
- Determine if capacity should be expressed in TB or PB
- Calculate the count (rounded up)

Example Usage :
```
// Form provides: bucket.capacityBytes = "1099511627776" (string)

const capacityBytes = parseCapacityBytes(bucket.capacityBytes); // 1099511627776 (number)
const limit = calculateStorageConsumptionLimit(capacityBytes);
// Returns: { kind: 'TB', count: 1 }

// Used in API request:
{
  storageConsumptionLimitKind: limit.kind,    // 'TB'
  storageConsumptionLimitCount: limit.count,  // 1
}
```

### DIFF
--- a/src/react/ISV/utils/capacityCalculations.ts
+++ b/src/react/ISV/utils/capacityCalculations.ts
@@ -1,0 +1,47 @@
+export const BYTES_IN_TIB = 1024 ** 4;
+export const BYTES_IN_PIB = 1024 ** 5;
+
+export type StorageConsumptionLimitKind = 'TB' | 'PB';
+
+export interface StorageConsumptionLimit {
+  kind: StorageConsumptionLimitKind;
+  count: number;
+}
+
+export function parseCapacityBytes(capacityBytes: string | number): number {
+  const parsed =
+    typeof capacityBytes === 'string'
+      ? parseInt(capacityBytes, 10)
+      : capacityBytes;
+  return isNaN(parsed) ? 0 : parsed;
+}
+
+export function calculateStorageConsumptionLimit(
+  capacityBytes: number,
+): StorageConsumptionLimit {
+  if (capacityBytes >= BYTES_IN_PIB) {
+    return {
+      kind: 'PB',
+      count: Math.ceil(capacityBytes / BYTES_IN_PIB),
+    };
+  }
+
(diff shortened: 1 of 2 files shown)


src/react/ISV/utils/capacityCalculations.ts
```diff
@@ -0,0 +1,47 @@
+export const BYTES_IN_TIB = 1024 ** 4;
+export const BYTES_IN_PIB = 1024 ** 5;
+
+export type StorageConsumptionLimitKind = 'TB' | 'PB';
```
ZiyangLinScality, line 4: TIB = 1024 ** 4
TB = 1000 ** 4

src/react/ISV/utils/capacityCalculations.ts
```diff
@@ -0,0 +1,47 @@
...
+  kind: StorageConsumptionLimitKind;
+  count: number;
+}
+
+export function parseCapacityBytes(capacityBytes: string | number): number {
+  const parsed =
+    typeof capacityBytes === 'string'
+      ? parseInt(capacityBytes, 10)
+      : capacityBytes;
+  return isNaN(parsed) ? 0 : parsed;
```
ZiyangLinScality: This doesn't handle negative values. Capacity should never be negative.

src/react/ISV/utils/capacityCalculations.ts
hervedombya: The Veeam API uses 'TB'/'PB' labels but expects binary calculations

src/react/ISV/utils/capacityCalculations.ts
hervedombya: The form validates capacity with `min={1}` and `Joi .min(1)`, so negative values shouldn't occur in practice. However, I add the check for more robustness !

ZiyangLinScality (approved):
LGTM

PR #1042, opened 2026-01-21
refactor: update SingleMutationDef to use dynamic MutationLabel for improved label handling

ZiyangLinScality:
"Create policy" in the action list is not correct as the described action, when extending a policy. When the policy already exists, it should be “Update Policy”


hervedombya (approved):
LGTM

PR #1043, opened 2026-01-21
refactor(ISVSummary): simplify access key label logic for improved readability

ZiyangLinScality:
"Access key IDs" in the copy all : should not be plural if only one


hervedombya (approved):
LGTM

PR #1050, opened 2026-02-02
Improvement/artesca 15745 iam role

JeanMarcMilletScality:
## Fix
Add ToastProvider to fix an error when trying to attach role with SelectAccountIAMRole:
image

### Attachment Tabs:
- Change order for user Tabs: Policy is now the default one
- Add EntityDisplayText to modify empty state text depending on entity name (group, policy, user, etc)

### SelectAccountIAMRole
- Add loading state when loading roles
- Use hiddenRoles to disabled already attached options 

See this doc for ref: https://docs.google.com/document/d/1H0X53PF4dWiOQy4FSr7zNq_XNei3Ifti6LCKxA2HTHo/edit?tab=t.0

### DIFF
--- a/src/react/ui-elements/SelectAccountIAMRole.tsx
+++ b/src/react/ui-elements/SelectAccountIAMRole.tsx
@@ -1,19 +1,24 @@
 import { Form, FormGroup, FormSection, ToastProvider, useToast } from '@scality/core-ui';
 import { Select } from '@scality/core-ui/dist/next';
-import { IAM } from 'aws-sdk';
-import { Bucket } from '@scality/data-browser-library';
-import { PropsWithChildren, useState } from 'react';
+import type { Bucket } from '@scality/data-browser-library';
+import { ShellHooksProvider } from '@scality/module-federation';
+import type { IAM } from 'aws-sdk';
+import { type PropsWithChildren, useState } from 'react';
 import { useMutation, useQuery, useQueryClient } from 'react-query';
 import { useParams } from 'react-router';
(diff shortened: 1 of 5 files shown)


src/react/ui-elements/SelectAccountIAMRole.tsx
```diff
@@ -299,19 +279,31 @@ const SelectAccountIAMRoleWithAccount = (
...
-                  <Select.Option key={`${role.RoleName}`} value={role.RoleName}>
-                    {role.RoleName}
-                  </Select.Option>
-                ))}
+                {
+                  roleQueryData.isLoading || !assumedRoleAccountMatchSelectedAccount
+                    ? <Select.Option key="loading" value="loading" disabled>Loading...</Select.Option>
+                    : roles.map((role) => {
+
+                      if (hideAccountRoles.find(hideRole => hideRole.accountName === accountName && hideRole.roleName === role.RoleName)) {
```
JBWatenbergScality: I'm not sure we put only role already attached in hideAccountRoles. I think we also put default seeds uninteresting for the user but I might be wrong to double check. Those default seeds we would really want to hide them

JeanMarcMilletScality: The roles in hideAccountRoles are the already attached (+ the pending ones with previous workflow).
Internal roles that we don't want to display are hidden thanks to `filterOutInternalRoles`

JBWatenbergScality (approved):

PR #1054, opened 2026-02-09
Bugfix/artesca 16828 truststore link

JeanMarcMilletScality:
Add check to display link only if Trust store is available (only when metalk8s is also deployed)
Fix link by using base path

### DIFF
--- a/src/react/truststore/EndpointInfoMessage.tsx
+++ b/src/react/truststore/EndpointInfoMessage.tsx
@@ -1,9 +1,13 @@
 import { InfoMessage, spacing } from '@scality/core-ui';
 import { Text } from '@scality/core-ui/dist/components/text/Text.component';
-import { LOCATION_EDITOR_FORCED_LABEL_WIDTH } from '../locations/LocationEditor';
 import { Box } from '@scality/core-ui/dist/next';
+import { LOCATION_EDITOR_FORCED_LABEL_WIDTH } from '../locations/LocationEditor';
+import { useConfig, useDeployedMetalk8sInstances } from '../next-architecture/ui/ConfigProvider';
 
 export const EndpointInfoMessage = ({ hasMargin }: { hasMargin?: boolean }) => {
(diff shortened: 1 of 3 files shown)


MonPote (approved):

src/react/utils/testUtil.tsx
```diff
@@ -606,45 +606,45 @@ const DataServiceProvider = ({ children }) => {
...
-                </DataServiceProvider>
-              </ToastProvider>
-            </MemoryRouter>
-          </ShellHooksProvider>
-        </ThemeProvider>
-      </QueryClientProvider>
-    );
-  };
+    ({ children }: { children: ReactNode }) => {
+      return (
```
hervedombya, line 610: I feel that there is formatter issue here, please check with biome that the file is correctly formatted

src/react/truststore/EndpointInfoMessage.tsx
```diff
@@ -1,9 +1,13 @@
...
 import { Text } from '@scality/core-ui/dist/components/text/Text.component';
-import { LOCATION_EDITOR_FORCED_LABEL_WIDTH } from '../locations/LocationEditor';
 import { Box } from '@scality/core-ui/dist/next';
+import { LOCATION_EDITOR_FORCED_LABEL_WIDTH } from '../locations/LocationEditor';
+import { useConfig, useDeployedMetalk8sInstances } from '../next-architecture/ui/ConfigProvider';
 
 export const EndpointInfoMessage = ({ hasMargin }: { hasMargin?: boolean }) => {
+  const { basePath } = useConfig();
+  const metalK8sInstances = useDeployedMetalk8sInstances();
+  const isMetalK8sEnabled = metalK8sInstances.length > 0;
```
hervedombya, line 10: nit: This can be extracted into a function to avoid duplicating this logic.

ChengYanJin (commented):
Just wondering should we hide the link when it's not available ?

PR #1058, opened 2026-02-12
feat(ISV): mark Veeam repository creation as optional mutation step

hervedombya:
## Summary

This PR makes the Veeam Repository auto-creation step a **graceful, optional mutation** in the ISV assistant. When the `createVeeamRepository` API call fails (e.g. Veeam server unreachable, timeout), the assistant no longer blocks the user — instead it shows a warning banner and lets them continue to manual configuration.

- **Veeam `createVeeamRepo` marked as `optional: true`** — failure no longer prevents the user from completing the wizard
- **Warning banner on optional failure** — displays "The automatic Veeam Repository creation was unsuccessful. You may continue to configure it manually."
- **"Continue to manual configuration" button** — replaces the standard "Continue" label when optional steps fail, making the fallback path explicit
- **Per-bucket failure messages** — each failed repository creation shows which bucket failed specifically
- **Summary fallback** — when auto-create fails, the summary shows the standard credentials view (access key, endpoint, etc.) instead of the success-only `VeeamRepositorySummary`

### How

The optional mutation infrastructure was wired in previous work (Tasks 1–4). This PR applies it to the Veeam VBR platform config:

- `veeam-vbr.tsx`: adds `optional`, `failureMessage`, `continueWithOptionalFailuresLabel`, and `defaultOptionalFailureMessage` to the platform definition; updates `customRender` to fall back to the default summary on optional failures
- `ISVApplyActions.test.tsx`: adjusts tests to account for VeeamVBRPlatform now carrying optional failure config

### Deviations from plan

The `defaultOptionalFailureMessage` type was changed from `string` to `ReactNode` (not in the original plan) to have more support.

### DIFF
--- a/src/react/ISV/engine/types.ts
+++ b/src/react/ISV/engine/types.ts
@@ -218,7 +218,7 @@ export type SingleMutationDef = {
   when?: (form: FormData, ctx: FullContext) => boolean;
   variables: (form: FormData, prev: PreviousResults, ctx: FullContext) => Record<string, unknown>;
   optional?: boolean;
-  failureMessage?: string | ((form: FormData, ctx: FullContext) => string);
+  failureMessage?: ReactNode | ((form: FormData, ctx: FullContext) => ReactNode);
 };
 
 export type PerBucketStep = {
@@ -228,7 +228,7 @@ export type PerBucketStep = {
   when?: (form: FormData, bucket: BucketItem, ctx: FullContext) => boolean;
   variables: (form: FormData, bucket: BucketItem, prev: PreviousResults, ctx: FullContext) => Record<string, unknown>;
   optional?: boolean;
-  failureMessage?: string | ((form: FormData, bucket: BucketItem, ctx: FullContext) => string);
+  failureMessage?: ReactNode | ((form: FormData, bucket: BucketItem, ctx: FullContext) => ReactNode);
 };
 
 export type LoopMutationDef = {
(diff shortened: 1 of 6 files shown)


src/react/ISV/engine/types.ts
ZiyangLinScality: Safe at runtime (string => ReactNode), but misleading — ReactNode will try returning JSX from failureMessage and hit the narrower type. Either widen failureMessage too, or add a brief comment explaining the intentional gap.

src/react/ISV/platforms/veeam-vbr.tsx
```diff
@@ -281,11 +239,9 @@ export const VeeamVBRPlatform = definePlatform({
     bucketBanner: <VeeamBucketBanner />,
     immutabilityLabel: 'Immutable Backup',
     immutabilityHelpText: (enabled: boolean) =>
-      enabled
-        ? 'Ensure "Make recent backups immutable" is checked when configuring the bucket in Veeam.'
-        : undefined,
-    customRender: ({ formData, onFinish, renderDefault }) => {
-      if (!formData.autoCreateRepository) {
+      enabled ? 'Ensure "Make recent backups immutable" is checked when configuring the bucket in Veeam.' : undefined,
+    customRender: ({ formData, onFinish, optionalFailures, renderDefault }) => {
```
ZiyangLinScality, line 243: need test coverage. customRender is a pure function

ZiyangLinScality (approved):

PR #1081, opened 2026-02-23
Auto-scroll ISV action tracking table to latest entry

hervedombya:
## Summary

- When the ISV assistant runs automatic configuration, the action tracking table now auto-scrolls to show the most recent entry as soon as the content overflows the visible area.
- The scroll happens once — after that, the user has full control over scrolling with no interference.

## Approach

A reusable `useAutoScrollOnce` hook observes the scroll container for DOM changes and scrolls to the bottom on first overflow, then stops observing. This keeps the component clean and the behavior self-contained, with no external dependency to manage.

## What changed from the initial plan

The initial ticket described auto-focus/auto-scroll behavior. During review, we chose a one-time scroll strategy instead of continuous auto-scroll, to avoid blocking the user from reviewing previous entries.

## Reference

[ZKUI-449](https://scality.atlassian.net/browse/ZKUI-449)

[ZKUI-449]: https://scality.atlassian.net/browse/ZKUI-449?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ZiyangLinScality (approved):
LGTM

PR #1085, opened 2026-02-25
Fix Veeam capacity display and edit update issues

ZiyangLinScality:
## Summary

- Fix AWS SDK v3 `ReadableStream` Body parsing — the original code used `.toString()` which returned `"[object ReadableStream]"`, causing capacity to always display as `0B`
- Fix form values not syncing when `maxCapacity` prop updates asynchronously (react-hook-form `defaultValues` only applies on initial mount)
- Fix UI not refreshing after editing capacity — replaced refetch-based approach with optimistic update + module-level `capacityCache` to handle `ReadableStream` single-read limitation
- Fix capacity showing `0B` after navigating away and back — `capacityCache` persists parsed values across component unmount/remount cycles

## Changes

**VeeamCapacityOverviewRow.tsx**
- Added `readBodyAsString()` to handle SDK v3 `ReadableStream.transformToString()` 
- Added `parseCapacityFromXml()` for clean XML parsing
- Added module-level `capacityCache` to persist parsed capacity across navigation
- Set `gcTime: 0` to prevent React Query from caching consumed streams
- Optimistic update via `onCapacityUpdated={handleCapacityUpdated}` instead of `refetch`

**VeeamCapacityModal.tsx**
- Added `onCapacityUpdated` callback prop to pass new capacity bytes to parent
- Added `useEffect` + `reset()` to sync form values when `maxCapacity` changes
- Extracted `capacityBytes` variable to avoid duplicate `getCapacityBytes()` calls

**Tests**
- Fixed test title grammar and accuracy
- Added test coverage for `onCapacityUpdated` callback
- Wrapped async capacity assertion in `waitFor`

## Test plan

- [x] 13 tests pass (`npm test -- src/react/ISV/components/veeam/`)
- [x] Manual: Edit capacity → verify UI updates immediately after save
- [x] Manual: Edit capacity → navigate away → navigate back → verify correct value displayed
- [x] Manual: First load of bucket overview → verify capacity displays correctly (not 0B)

### DIFF
--- a/src/react/ISV/components/veeam/VeeamCapacityOverviewRow.tsx
+++ b/src/react/ISV/components/veeam/VeeamCapacityOverviewRow.tsx
@@ -1,16 +1,28 @@
-import {
-  useGetBucketTagging,
-  useGetObject,
-} from '@scality/data-browser-library';
+import { PrettyBytes } from '@scality/core-ui';
+import { useGetBucketTagging, useGetObject } from '@scality/data-browser-library';
+import { useCallback, useEffect, useState } from 'react';
 import * as T from '../../../ui-elements/TableKeyValue2';
-import { VeeamCapacityModal } from './VeeamCapacityModal';
 import {
   BUCKET_TAG_APPLICATION,
   BUCKET_TAG_VEEAM_APPLICATION,
   VEEAM_OBJECT_KEY,
   VeeamApplicationType,
 } from '../../constants';
-import { PrettyBytes } from '@scality/core-ui';
+import { VeeamCapacityModal } from './VeeamCapacityModal';
+
+function parseCapacityFromXml(xml: string): number {
+  const text = new DOMParser().parseFromString(xml, 'application/xml').querySelector('Capacity')?.textContent;
+  return text ? parseFloat(text) : 0;
+}
+
+async function readBodyAsString(body: NonNullable<unknown>): Promise<string> {
+  if (typeof (body as Record<string, unknown>).transformToString === 'function') {
+    return (body as { transformToString: () => Promise<string> }).transformToString();
+  }
+  return String(body);
+}
+
+const capacityCache = new Map<string, number>();
 
 const VeeamCapacityContent = ({ bucketName }: { bucketName: string }) => {
   const { data: taggingData, status: taggingStatus } = useGetBucketTagging({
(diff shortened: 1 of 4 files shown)


src/react/ISV/components/veeam/VeeamCapacityOverviewRow.tsx
```diff
@@ -1,97 +1,97 @@
...
 } from '../../constants';
-import { PrettyBytes } from '@scality/core-ui';
+import { VeeamCapacityModal } from './VeeamCapacityModal';
+
+function parseCapacityFromXml(xml: string): number {
+  const text = new DOMParser().parseFromString(xml, 'application/xml').querySelector('Capacity')?.textContent;
+  return text ? parseFloat(text) : 0;
+}
+
+async function readBodyAsString(body: NonNullable<unknown>): Promise<string> {
```
hervedombya, line 18: The typing casts is fragile. Do the SDK v3 types are available ?

src/react/ISV/components/veeam/VeeamCapacityOverviewRow.tsx
```diff
@@ -1,97 +1,97 @@
...
-    {
-      Bucket: bucketName,
-      Key: VEEAM_OBJECT_KEY,
-    },
-    {
-      enabled: taggingStatus === 'success' && isSOSAPIEnabled,
-      retry: false,
-    },
+    { Bucket: bucketName, Key: VEEAM_OBJECT_KEY },
+    { enabled: taggingStatus === 'success' && isSOSAPIEnabled, retry: false, gcTime: 0 },
```
hervedombya: `gcTime: 0` is a workaround for the stream consumption issue. Now that the body is properly parsed and cached in state, this might cause unnecessary re-fetching on component
  remount. Worth verifying the interaction with `capacityCache`.

src/react/ISV/components/veeam/VeeamCapacityOverviewRow.tsx
ZiyangLinScality: The project uses aws-sdk v2, not v3 — SDK v3 types (@smithy/types, @aws-sdk/client-s3) aren't in the dependency tree. They're only internal to data-browser-library. Duck typing bridges the SDK v2/v3 gap for now. I'd prefer to address this properly when we upgrade to AWS SDK v3.

hervedombya: Ok, to address when we gonna update aws-sdk here then

hervedombya (approved):

PR #1108, opened 2026-03-16
refactor: redesign credential lifecycle with provider pattern and proactive refresh

ZiyangLinScality:
## Summary

- **Replace `useMutation` with `useQuery` for AssumeRole**: Leverages React Query's built-in caching, `refetchOnWindowFocus`, and `staleTime` instead of manually managing mutation state and `useEffect` triggers.
- **Introduce credential provider pattern**: Instead of passing static credentials to `DataBrowserProvider`, passes an async `credentialProvider` function. The AWS SDK calls this before each request, enabling transparent credential refresh without re-rendering the component tree.
- **Add proactive credential refresh with 2-minute buffer**: Credentials are refreshed before they expire (`REFRESH_BUFFER_MS = 2min`), preventing `ExpiredToken` errors after idle periods. A deduplication mechanism (`refreshPromiseRef`) ensures only one refresh request is in-flight at a time.
- **Simplify `setRole` / `setRolePromise`**: Role switching now leverages `queryClient.fetchQuery` instead of manual mutation + callback wiring.

## Changed Files

| File | Change |
|------|--------|
| `DataServiceRoleProvider.tsx` | Core refactor: `useMutation` → `useQuery`, static credentials → credential provider, added proactive refresh logic |
| `DataServiceRoleProvider.test.tsx` | New integration tests covering render, role context, and STS credential flow |
| `testUtil.tsx` | Simplified test wrapper to remove unused `useAssumeRoleQuery` / `useMutation` dependencies |

## Motivation

When the browser tab is idle and STS temporary credentials expire, subsequent S3 requests fail with `ExpiredToken` errors. The previous design relied on `useMutation` triggered by `useEffect`, which only refreshed credentials on role or token changes — not on expiration. The new credential provider pattern lets the AWS SDK resolve fresh credentials on demand, and the 2-minute pre-expiry buffer ensures seamless operation after idle periods.

## Test Plan

- [x] Integration tests for `DataServiceRoleProvider` added
- [x] Manual: verify S3 operations work after leaving tab idle past credential expiration
- [x] Manual: verify role switching still works correctly
- [x] Manual: verify `refetchOnWindowFocus` triggers credential refresh when returning to tab

### DIFF
--- a/src/react/DataServiceRoleProvider.tsx
+++ b/src/react/DataServiceRoleProvider.tsx
@@ -4,7 +4,7 @@ import type { AWSError, STS } from 'aws-sdk';
 import type { PromiseResult } from 'aws-sdk/lib/request';
 import { createContext, type JSX, useContext, useEffect, useMemo, useState } from 'react';
 import { flushSync } from 'react-dom';
-import { useMutation } from 'react-query';
+import { useQuery, useQueryClient } from 'react-query';
 import { useParams } from 'react-router';
 import { useTheme } from 'styled-components';
 import STSClient from '../js/STSClient';
@@ -195,7 +195,7 @@ const DataServiceRoleProvider = ({
   const accountName = params?.accountName;
   const theme = useTheme();
 
-  const { iamInternalFQDN, s3InternalFQDN, zenkoEndpoint, iamEndpoint } = useConfig();
+  const { iamInternalFQDN, s3InternalFQDN, zenkoEndpoint, iamEndpoint, stsEndpoint } = useConfig();
 
   useEffect(() => {
     initializeAWSSigner({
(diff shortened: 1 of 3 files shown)


ChengYanJin (approved):

src/react/DataServiceRoleProvider.tsx
```diff
@@ -241,53 +242,91 @@ const DataServiceRoleProvider = ({
...
+  const assumeRoleQueryRef = useRef(assumeRoleQuery);
+  assumeRoleQueryRef.current = assumeRoleQuery;
+
+  const refreshPromiseRef = useRef<Promise<S3Credentials> | null>(null);
+  const refreshCooldownUntilRef = useRef<number>(0);
+  const REFRESH_COOLDOWN_MS = 30 * 1000;
+
+  const credentialProvider = useCallback(async (): Promise<S3Credentials> => {
+    const expiration = assumeRoleQueryRef.current.data?.Credentials?.Expiration;
+    const REFRESH_BUFFER_MS = 2 * 60 * 1000;
```
JBWatenbergScality, line 263: Why 2 minutes ? I think assumeRole credentials are valid for 10 minutes by default (to be double checked). Can we use instead expiry information from those credentials if the AWS standard gives us the expiry date and renew something like 1 minute before expiry as we do for the access token ?

ZiyangLinScality: 1. AssumeRole credentials are 15 minutes, not 10 — we set DurationSeconds: 900 in STSClient.ts.
 2. We already use the expiry date from the STS response. The 2-minute REFRESH_BUFFER_MS is not a polling interval — it's a buffer before expiry. So for 15-min credentials, refresh triggers at ~13 min.
 3. 2 minutes gives more safety margin than 1 minute for network latency / retries, but happy to reduce to 1 min if we want consistency with the access token pattern.

PR #1112, opened 2026-03-23
fix(ZKUI-475): Only display Storage Manager role when it supersedes others

ZiyangLinScality:
## Summary

- When a user has `storage-manager-role` for an account, all other IAM roles are now hidden in the Select Role modal
- Storage Manager provides a superset of permissions over every other role, so displaying lower-permission roles alongside it is redundant and confusing
- Applied filtering in both role selection components:
  - `AccountRoleSelectButtonAndModal` (table-based modal)
  - `SelectAccountIAMRole` (dropdown-based selector)

## Changes

- **`AccountRoleSelectButtonAndModal.tsx`**: Added per-account check — if `storage-manager-role` exists among the account's roles, only that role is included in the list
- **`SelectAccountIAMRole.tsx`**: Added post-filter step after the existing `filterOutInternalRoles` logic — if `storage-manager-role` is in the fetched roles, only it is kept
- **`SelectAccountIAMRole.test.tsx`**: Updated shared mock data and added a dedicated test verifying that only `storage-manager-role` is displayed when present alongside other roles

## Test plan

- [x] Existing tests pass (11 passed, 1 previously skipped)
- [x] New test: verifies `data-consumer-role` and `storage-account-owner-role` are hidden when `storage-manager-role` is present
- [x] Manual verification: open Select Role modal with a Storage Manager account and confirm only `storage-manager-role` appears

### DIFF
--- a/src/react/account/AccountRoleSelectButtonAndModal.tsx
+++ b/src/react/account/AccountRoleSelectButtonAndModal.tsx
@@ -107,23 +107,19 @@ export function AccountRoleSelectButtonAndModal({
     return (
       accounts?.flatMap((account) => {
         const accountName = account.Name;
-        const hasStorageManager = account.Roles.some(
-          (role) => regexArn.exec(role.Arn)?.groups?.['name'] === STORAGE_MANAGER_ROLE,
-        );
-        const rolesToDisplay = hasStorageManager
-          ? account.Roles.filter((role) => regexArn.exec(role.Arn)?.groups?.['name'] === STORAGE_MANAGER_ROLE)
-          : account.Roles;
-        return rolesToDisplay.map((role) => {
+        const parsedRoles = account.Roles.map((role) => {
           const parsedArn = regexArn.exec(role.Arn);
-          const rolePath = parsedArn?.groups['path'] || '';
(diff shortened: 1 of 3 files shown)


src/react/account/AccountRoleSelectButtonAndModal.tsx
```diff
@@ -107,7 +107,13 @@ export function AccountRoleSelectButtonAndModal({
     return (
       accounts?.flatMap((account) => {
         const accountName = account.Name;
-        return account.Roles.map((role) => {
+        const hasStorageManager = account.Roles.some(
+          (role) => regexArn.exec(role.Arn)?.groups?.['name'] === STORAGE_MANAGER_ROLE,
+        );
+        const rolesToDisplay = hasStorageManager
+          ? account.Roles.filter((role) => regexArn.exec(role.Arn)?.groups?.['name'] === STORAGE_MANAGER_ROLE)
```
hervedombya: minor: the regex is executed twice for each role (once to check h`asStorageManager`, once to filter). A minor optimization would be to extract role names once

hervedombya (approved):

PR #1179, opened 2026-04-30
ZKUI-497: Share ISV Connector Modal via Module Federation

JeanMarcMilletScality:
## Context

The new Overview page (ARTESCA-14397) needs to open the ISV Connector selection dialog from outside zenko-ui. The existing `ISVModal` only works inside zenko-ui (uses local React Router navigation). This PR exposes a federated wrapper — `ISVConnectorModal` — following the same pattern as `SelectAccountIAMRole`, so any UI can import and render it with correct cross-app navigation.

## Summary of changes

- **`ISVConnectorModal.tsx`** (new): federated wrapper that accepts `{ isOpen, setIsOpen, shellHooks, shellAlerts }`. Owns `selectedISV` state, computes the target view from `selectedISV` (mirroring `WelcomeModal`'s `zenkoUIConfigurationView`), and navigates via `openLink` (cross-app) or `HistoryPushEvent` (already in zenko-ui).
- **`ISVModal.tsx`**: fix `return null` instead of `return <></>` (biome lint).
- **`rspack.config.ts`**: expose `./ISVConnectorModal` in the Module Federation config.

### DIFF
--- a/src/react/ISV/components/Modal/ISVConnectorModal.tsx
+++ b/src/react/ISV/components/Modal/ISVConnectorModal.tsx
@@ -1,0 +1,105 @@
+import { Icon, Modal, Stack, Text, Wrap } from '@scality/core-ui';
+import { Button } from '@scality/core-ui/dist/next';
+import { ShellHooksProvider, useShellHooks } from '@scality/module-federation';
+import { useMemo, useState } from 'react';
+import type { ShellAlerts, ShellHooks } from 'shell/compiled-types/src/hooks/useShellHooks';
+import styled from 'styled-components';
+import type { ISVCardConfig } from '../../types';
+import { ArtescaLogo } from '../ArtescaLogo';
+import { ISVModalContent } from './ISVModal';
+
+const CustomModal = styled(Modal)`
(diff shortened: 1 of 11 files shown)


src/react/ISV/components/Modal/ISVConnectorModal.tsx
```diff
@@ -0,0 +1,105 @@
...
+`;
+
+type ISVConnectorModalProps = {
+  isOpen: boolean;
+  setIsOpen: (open: boolean) => void;
+  shellHooks: ShellHooks;
+  shellAlerts: ShellAlerts;
+};
+
+const ISVConnectorModalInternal = ({
```
ChengYanJin, line 19: There are a lot of duplicated code with WelcomeModal. Should we extract some common logic ?

JeanMarcMilletScality: Yes I agree. We should either share the logic or should we use only one share component ? And have a variant WelcomeModal using condition for the isOpen (making it optional?)

ChengYanJin (approved):
LGTM!

PR #1109, opened 2026-03-17
feat(ZKUI-472): introduce WebMCP tools for AI agent automation

JBWatenbergScality:
## Summary

Introduces WebMCP tools for AI agent automation in zenko-ui, exposed via the shell-ui \`MCPTools\` federated module contract.

- **mcpTools/index.ts**: exports the \`tools\` array consumed by shell-ui's \`MCPRegistrar\`
- **ToolContext / types.ts**: mirrors shell-ui's \`ToolContext\` — includes \`getToken\`, \`userData\`, and \`selfConfiguration\`; \`authRequired\` removed (shell-ui always injects auth context)
- **Tools** (all receive \`getToken\` and \`userData\` via context):
  - \`createAccount\` — create a new Zenko account
  - \`getAssumableRoles\` — list IAM roles the current user can assume
  - \`getCredentialsInstructions\` — get STS credentials for a role
  - \`getAWSCLIS3Instructions\` — generate AWS CLI S3 commands
  - \`getAWSCLIIamInstructions\` — generate AWS CLI IAM commands
  - \`getS3Actions\` — list available S3 actions
  - \`getIAMActions\` — list available IAM actions
  - \`executeS3Action\` — execute an S3 SDK action
  - \`executeIAMAction\` — execute an IAM SDK action

PR #1185, opened 2026-05-08
feat(ZKUI-472): WebMCP tools for AI agent automation

JBWatenbergScality:
## Summary

Introduces the WebMCP tool surface for zenko-ui exposed via the shell-ui `MCPTools` federated module contract (`createTools` factory). Covers both the zenko-ui scoped tools and the wiring needed to make the upstream `@scality/data-browser-library` S3 tools work against a Zenko deployment.

Closes ZKUI-472.

### Zenko-scoped MCP tools (`src/react/mcpTools/tools/`)

All tools receive shell-ui's `ToolContext` (`getToken`, `userData`, `selfConfiguration`, `queryClient`) baked into their `execute` closure by the `createTools` factory.

- `createAccount` — creates a new Zenko account; invalidates `['accounts']` on success
- `getAssumableRoles` *(read-only)* — STS `AssumeRoleWithWebIdentity` role list; populates `accountsCache` so other tools can resolve `roleArn → accountName`
- `getCredentialsInstructions` *(read-only)* — emits the ready-to-run `aws sts assume-role-with-web-identity` command + env exports
- `getAWSCLIS3Instructions` / `getAWSCLIIamInstructions` *(read-only)* — emit ready-to-run `aws s3` / `aws iam` commands with the right endpoint flags

Read-only tools carry `annotations: { readOnlyHint: true }` so MCP clients can render them as safe-to-auto-run.

> The executable IAM surface (`getIAMActions`, `executeIAMAction`) is intentionally out of scope here and will land in a dedicated follow-up PR.

### Wrapping data-browser S3 tools for Zenko (`createZenkoS3Tools.ts`)

Each upstream data-browser S3 tool is wrapped so it works against a Zenko deployment:

- **Host `basePath` derivation** — composes the host prefix from `selfConfiguration.basePath` directly (the registry advertises the same value; combining both produced `/data/data/...` previously).
- **Account name resolution** — explicit LLM arg → URL → `accountsCache` by `roleArn` → sole-account fallback. Each step is `console.debug`-logged.
- **Two-stage proxy SigV4** — `getS3Config()` returns `endpoint = https://<s3InternalFQDN>` plus a `proxy` block so the lib's pre-signing middleware strips `/zenko/s3` and rewrites `Host` to the internal FQDN before SigV4 — mirrors `initializeAWSSigner` in `src/react/utils/index.ts` for the v2 SDK side of the app.
- **`navigateToRoute` schema** gains an optional `accountName` so the LLM can navigate from `/` or `/data/accounts` (no account in URL).
- **`normalizeStringifiedSchemaArgs`** — walks `tool.inputSchema.properties`, resolves `$ref`s, then `JSON.parse`s any string passed for an `object`/`array`-typed property. The LLM consistently ships nested params (most often `Metadata`) as JSON-encoded strings under `dict[str, Any]` structured output; without this they reached the SDK as strings and silently dropped on the wire.
- **`accountsCache.ts`** / **`stsCredentialCache.ts`** — module-level caches so wrappers can map `roleArn → accountName` and reuse STS credentials within their TTL.

### Cache invalidation

Shell-ui injects its shared `QueryClient` via `context.queryClient`. Each tool emits **targeted** invalidations at the end of its `execute` (e.g. `createAccount` invalidates `['accounts']`). No outer wrapper blanket-invalidates after every call — that approach blew away the auth queries in an earlier iteration and left the UI empty. Invalidation runs on the data-browser-library v5 QueryClient, not shell-ui v3.

### Dependency bump

- `@scality/data-browser-library`: 1.1.20 → **1.1.22**, which ships the upstream equivalents of the `patch-package` patches we were carrying (BucketList responsive header, `autoNavigate`, two-stage SigV4 middleware, generated MCP tool fixes). The `patches/@scality+data-browser-library+1.1.20.patch` is dropped.

### Other plumbing

- `types.ts` — `ToolContext`, `ZenkoSelfConfiguration` (`managementEndpoint`, `iamEndpoint`, `stsEndpoint`, `zenkoEndpoint`, `s3InternalFQDN`), `buildZenkoContext`
- `public/.well-known/micro-app-configuration` + `rspack.config.ts` — exposes `MCPTools` as a federated module so shell-ui's `MCPRegistrar` picks it up
- `src/js/accountActions.ts` — `createAccountCore` extracted from `useCreateAccountMutation` so the tool can reuse the same code path

## Test plan

- [ ] First-try tool chain (`getAssumableRoles → assumeRoleWithWebIdentity → headObject → autoNavigate`) lands the user on `/data/accounts/<account>/buckets/<bucket>/objects?…`
- [ ] S3 calls reach `/zenko/s3/<bucket>/<key>` with a 200 (no `SignatureDoesNotMatch`)
- [ ] `putObject` with custom metadata sets `x-amz-meta-*` headers (Network tab confirms; `headObject` reflects)
- [ ] `navigateToRoute` from `/` works when the LLM passes `accountName` or when only one assumable account is cached
- [ ] `createAccount` triggers a refresh of the accounts list in the UI (no full reload needed)
- [ ] Read-only tools surface `readOnlyHint: true` in the MCP client tool list
- [ ] `tsc --noemit` clean
- [ ] Existing zenko-ui flows unaffected (the v2 SDK `AWS.Signers.V4` monkey-patch still owns non-MCP calls)

Supersedes #1109.


src/react/mcpTools/tools/createAccountTool.ts
```diff
@@ -0,0 +1,43 @@
+import { createAccountCore } from '../../../js/accountActions';
+import { buildZenkoContext, extractInstanceId, ToolContext } from '../types';
+
+export const createAccountTool = {
+  name: 'createAccount',
+  description:
+    'Creates a new ARTESCA account with a unique name and a root email address. Returns the created account ID.',
```
JBWatenbergScality:
```suggestion
    'Creates a new account with a unique name and a root email address. Returns the created account ID.',
```

src/react/mcpTools/tools/getAWSCLIIamInstructionsTool.ts
```diff
@@ -0,0 +1,51 @@
+import { buildZenkoContext, ToolContext } from '../types';
+
+export const getAWSCLIIamInstructionsTool = {
+  name: 'getAWSCLIIamInstructions',
+  description:
+    'Returns ready-to-run AWS CLI commands to manage IAM users, access keys, and policies on ARTESCA. ' +
```
JBWatenbergScality:
```suggestion
    'Returns ready-to-run AWS CLI commands to manage IAM users, access keys, and policies. ' +
```

src/react/mcpTools/tools/getAWSCLIS3InstructionsTool.ts
```diff
@@ -0,0 +1,47 @@
+import { buildZenkoContext, ToolContext } from '../types';
+
+export const getAWSCLIS3InstructionsTool = {
+  name: 'getAWSCLIS3Instructions',
+  description:
+    'Returns ready-to-run AWS CLI commands to interact with ARTESCA S3-compatible object storage. ' +
```
JBWatenbergScality:
```suggestion
    'Returns ready-to-run AWS CLI commands to interact with S3-compatible object storage. ' +
```

src/react/mcpTools/createZenkoS3Tools.ts
```diff
@@ -0,0 +1,356 @@
...
+  const basePath = `${zenkoBasePath}${accountSegment}`;
+  console.debug(
+    '[deriveHostBasePath] zenkoBasePath=%o accountName=%o (explicit=%o, fromUrl=%o, fromCache=%o, fromSole=%o) → basePath=%o',
+    zenkoBasePath, accountName, explicitAccountName, fromUrl?.[1], fromCache, fromSole, basePath,
+  );
+  return basePath;
+}
+
+/**
+ * Builds the full set of data-browser S3 tools adapted for the Zenko/ARTESCA context:
```
JBWatenbergScality:
```suggestion
 * Builds the full set of data-browser S3 tools adapted for the Zenkocontext:
```

src/react/mcpTools/createZenkoS3Tools.ts
```diff
@@ -0,0 +1,356 @@
...
+      // shell catch-all would redirect to `/data/accounts`.
+      const inputSchema = tool.name === 'navigateToRoute'
+        ? {
+            ...tool.inputSchema,
+            properties: {
+              ...(tool.inputSchema.properties as Record<string, unknown>),
+              accountName: {
+                type: 'string',
+                description:
+                  'ARTESCA account whose data-browser the navigation targets ' +
```
JBWatenbergScality:
```suggestion
                  'Account whose data-browser the navigation targets ' +
```

src/react/mcpTools/types.ts
```diff
@@ -0,0 +1,69 @@
...
+   */
+  queryClient?: QueryClient;
+};
+
+export type ZenkoSelfConfiguration = {
+  managementEndpoint: string;
+  iamEndpoint: string;
+  stsEndpoint: string;
+  zenkoEndpoint: string;
+  // Internal cluster-local hostname the ARTESCA ingress rewrites Host to before
```
JBWatenbergScality:
```suggestion
  // Internal cluster-local hostname the ingress rewrites Host to before
```

src/react/mcpTools/stsCredentialCache.ts
```diff
@@ -0,0 +1,43 @@
...
+
+type CachedCredentials = {
+  AccessKeyId: string;
+  SecretAccessKey: string;
+  SessionToken: string;
+  Expiration: Date;
+};
+
+// Module-level cache: roleArn → credentials
+const cache = new Map<string, CachedCredentials>();
```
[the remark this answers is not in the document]
JBWatenbergScality: Concurent call shouldn't happen and cache miss on concurent would be a minor issue here, we can live with this implementation

src/react/mcpTools/types.ts
```diff
@@ -0,0 +1,69 @@
...
+  // forwarding /zenko/s3/* to cloudserver. Required for SigV4 to validate at
+  // the backend — see createZenkoS3Tools and the v2 monkey-patch in
+  // src/react/utils/index.ts (initializeAWSSigner).
+  s3InternalFQDN: string;
+};
+
+export type ZenkoToolContext = ToolContext & ZenkoSelfConfiguration;
+
+export function buildZenkoContext(context: ToolContext): ZenkoToolContext {
+  const cfg = context.selfConfiguration as ZenkoSelfConfiguration;
```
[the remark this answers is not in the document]
JBWatenbergScality: It is part of zenko self configuration and enforced by zenko-ui-operator, ignorable

src/react/mcpTools/createZenkoS3Tools.ts
```diff
@@ -0,0 +1,356 @@
...
+ *
+ * @param context  Shell-ui ToolContext (getToken, userData, selfConfiguration).
+ * @param navigate React Router navigate function provided by shell-ui's MCPRegistrar.
+ */
+export function createZenkoS3Tools(
+  context: ToolContext,
+  navigate: (path: string) => void,
+): MCPToolDefinition[] {
+  const ctx = buildZenkoContext(context);
+  const features = (ctx.selfConfiguration.features as string[] | undefined) ?? [];
```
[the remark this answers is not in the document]
JBWatenbergScality: Features changing at runtime won't happen they are distributed via the wbe finger which is loaded only once at app bootstrap by shell-ui

src/react/mcpTools/createZenkoS3Tools.ts
```diff
@@ -0,0 +1,267 @@
...
+import { getCredentials } from './stsCredentialCache';
+import { buildZenkoContext, type ToolContext } from './types';
+
+/**
+ * S3 region the chat-side DataBrowserProvider uses
+ * (see DataServiceRoleProvider.tsx → DEFAULT_REGION). Mirrored here so the
+ * tool-side queryKeyPrefix matches what the panel namespaces queries against;
+ * if the panel and tools disagree the invalidations would silently miss.
+ */
+const PANEL_REGION = 'us-east-1';
```
damiengillesscality, line 16: Not clear to me what it does, but I smell something that may cause tricky issues in other region, could you please dooble check that

damiengillesscality (approved):

src/react/mcpTools/createZenkoS3Tools.ts
JBWatenbergScality: It's expected. Our S3 implementation appears "single-regioned" under this one for S3 namespace. Which can be diffrent than the location constraint of a bucket. Anyway here it is ana expected hardcoded value at this level

PR #1217, opened 2026-05-29
improvement(ARTESCA-17610): remove useless Data Browser guard for storage-usage-consumer role

hervedombya:
## Summary

- Removes `DataBrowserGuard` component that blocked Data Browser routes for the `storage-usage-consumer` role
- Removes `isStorageUsageConsumerRole` sidebar hiding that hid the Data Browser link
- Removes `getIsStorageUsageConsumerRole()` helper (only used for the guard)
- Updates tests to reflect that the Data Browser is now accessible to all roles

## Why

The guard relied on reading the role ARN from localStorage, which caused a race condition on login/logout: the stale role ARN would briefly give the wrong result, causing the Data Browser to flash hidden or visible for the wrong user.

The data-browser library (bumped in ZKUI-477) now shows a proper "Permission denied" message when `ListBuckets` returns a 403, which is the correct place to handle this — it reacts to the actual API response rather than cached state.

## Test plan

- [ ] Log in as a `storage-usage-consumer` user — Data Browser sidebar entry is visible
- [ ] Navigate to Data Browser as `storage-usage-consumer` — "Permission denied" message is shown
- [ ] Log out and log in as a `storage-manager` user — Data Browser works normally with no flicker
- [ ] All Routes tests pass

### DIFF
--- a/src/react/Routes.test.tsx
+++ b/src/react/Routes.test.tsx
@@ -454,7 +454,7 @@ describe('Routes component', () => {
     });
   });
 
-  describe('Data Browser route guards', () => {
+  describe('Data Browser sidebar', () => {
     const STORAGE_USAGE_CONSUMER_ARN = 'arn:aws:iam::000000000000:role/scality-internal/storage-usage-consumer-role';
     const STORAGE_MANAGER_ARN = 'arn:aws:iam::000000000000:role/scality-internal/storage-manager-role';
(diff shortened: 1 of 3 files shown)


ChengYanJin (approved):

JBWatenbergScality (approved):
LGTM

PR #1238, opened 2026-07-15
feature(crr): add crrService type definitions

hervedombya:
Gives the frontend and reviewers the shared TypeScript vocabulary for the `crr-configurator` REST API. Without this, follow-up PRs would either re-derive the same shapes inline in every hook or import them from a runtime module that does not yet exist.

The types mirror the contract in the backend repo's [`openapi.yaml`](https://github.com/scality/crr-configurator/blob/main/openapi.yaml) — request bodies for `/verify` and `/replication-setups`, the discriminated `VerifyResponse`, the five NDJSON `SetupEvent` variants, and the RFC 7807 `Problem` shape with its stable `code` enum. Nothing here has runtime behaviour, so TypeScript itself validates the file; no tests to add.

Next step: a typed `fetch` helper that parses `application/problem+json` responses into a `ServiceError`, then the first react-query mutation hook (`useVerifyMutation`) that consumes both.

### DIFF
--- a/src/react/locations/CRRSetupWizard/api/types.ts
+++ b/src/react/locations/CRRSetupWizard/api/types.ts
@@ -1,0 +1,111 @@
+// Types for the crr-configurator REST API.
+// Source of truth: `openapi.yaml` in scality/crr-configurator.
+//
+// This file contains types only — no runtime code. The fetch helper
+// and react-query hooks that use these types live in follow-up
+// modules.
+
+export type HostAlias = { hostname: string; ip: string };
+
+export type DestinationConnection =
+  | {
+      mode: 'management-network';
+      baseUrl: string;
+      adminUser: string;
+      adminPassword: string;
+    }
+  | {
+      mode: 'data-network';
+      baseDomain: string;
+      s3Endpoint: string;
+      adminUser: string;
+      adminPassword: string;
+    };
+
+export type VerifyRequestBody = {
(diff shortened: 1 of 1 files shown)


src/react/locations/CRRSetupWizard/api/types.ts
```diff
@@ -0,0 +1,111 @@
...
+//
+// This file contains types only — no runtime code. The fetch helper
+// and react-query hooks that use these types live in follow-up
+// modules.
+
+export type HostAlias = { hostname: string; ip: string };
+
+export type DestinationConnection =
+  | {
+      mode: 'cluster-url';
```
JBWatenbergScality: I think management-network would be a better name. To align with official documentation terms https://documentation.scality.com/Artesca/4.2.3/installation/hardware-appliance/hw_network_consideration.html#project-networks

src/react/locations/CRRSetupWizard/api/types.ts
```diff
@@ -0,0 +1,111 @@
...
+      baseUrl: string;
+      adminUser: string;
+      adminPassword: string;
+    }
+  | {
+      mode: 'data-plane';
```
JBWatenbergScality:
```suggestion
      mode: 'data-network';
```

JBWatenbergScality (approved):

PR #1240, opened 2026-07-16
feature(crr): add crrFetch helper and useVerifyMutation

hervedombya:
Adds the frontend seam that the CRR Provisioning Wizard will use to reach the `crr-configurator` service. `postJSON` sends the wizard's Keycloak session cookie, throws a typed `ServiceError` when the service replies with RFC 7807 `problem+json`, and carries the `code` (plus `unresolvedHosts` on `DestinationDnsResolutionFailed`) so the wizard can branch on failure kind without re-parsing.

Wraps that helper in a `useVerifyMutation` react-query hook that yields the mode-specific `VerifyResponse` and surfaces `ServiceError` on failure. Both the helper and the hook are covered by MSW-based tests.

Next brick wires the same pattern to `POST /api/v1/replication-setups` via an NDJSON-reading `useReplicationSetupsStream`.


src/react/locations/CRRSetupWizard/api/useVerifyMutation.ts
```diff
@@ -0,0 +1,12 @@
+import { useMutation } from 'react-query';
+import { postJSON } from './crrFetch';
+import type { VerifyRequestBody, VerifyResponse } from './types';
+
+// react-query mutation that POSTs the wizard's verify payload to
+// `/crr-configurator/api/v1/verify` and yields the mode-specific
+// response. Errors surface as `Error` — cast to `ServiceError` at
+// the caller to read `.problem.code`.
+export const useVerifyMutation = () =>
```
JBWatenbergScality:
```suggestion
export const useCRRConfigurationVerifyMutation = () =>
```

JBWatenbergScality (approved):

PR #1241, opened 2026-07-16
feature(crr): add crrConfigurator client and react-query hooks for verify + setup

hervedombya:
Adds the two entry points the CRR Provisioning Wizard uses to talk to the `crr-configurator` service: a verify call to check a destination cluster, and a setup call that runs the provisioning chain and reports its progress step by step. Both are surfaced as React hooks and carry the caller's ARTESCA session token as a Bearer header.

The network layer is a small standalone client, tested end-to-end without any UI framework. The wizard hooks wrap that client and expose react-query state (loading / success / error / data); the setup hook additionally exposes the ongoing progress events and a way to cancel.

Deviation from the plan: originally scoped as a bespoke NDJSON streaming hook with its own state machine. Simplified during review to align on the same react-query shape as verify, with the network layer extracted alongside. The `Authorization: Bearer` switch was folded in as well, replacing a never-shipped cookie assumption.

Next: the wizard component itself, gated behind the CRR feature flag.


src/react/locations/CRRSetupWizard/hooks/useCRRConfigurationSetupMutation.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  const mutation = useMutation<SetupResult, Error, StartSetupBody>({
+    mutationFn: async (body) => {
+      abortRef.current?.abort();
+      const controller = new AbortController();
+      abortRef.current = controller;
+      setEvents([]);
+      const token = await getToken();
+      for await (const event of startSetup(body, { token, signal: controller.signal })) {
+        setEvents((prev) => [...prev, event]);
```
JBWatenbergScality: This mix up react state and react query cache, likely we should store events in react query cache it seems to me to make more sense. This discusion provides some way to do that https://github.com/TanStack/query/discussions/418

JBWatenbergScality (approved):

PR #1243, opened 2026-07-17
feature(crr): add wizard shell and Configure step with source/destination/replication sections

hervedombya:
Lands the CRR Provisioning Wizard's shell and its first step. Users open it from a `Start CRR Configuration` button on the Locations page (only visible when the `CRR_CONFIGURATOR` feature flag is enabled) and are presented with a single Configure page grouped into three sections: Source (create or reuse an account on the source cluster), Destination Connection (network mode, URL or base domain + S3 endpoint, admin credentials, PEM certificate, destination account name), and Replication (an optional replication rule with source bucket, target bucket and prefix). A Check Connection button inside the Destination section validates the destination on demand and shows a success or error toast tied to the ARTESCA problem code. Continue silently re-verifies unless the last verify already covered the exact same connection values, and blocks progression with an error toast if that verify fails; a successful advance is transparent, as expected in a wizard flow. Apply actions and Summary land as placeholders and are filled in by the follow-up brick.

Under the hood, the Configure step is composed of three small section components that share form state via `FormProvider`, wrapping a strict Joi schema with conditional branches on the connection mode and the replication toggle. The PEM certificate input is a reusable `CertificateSection` copied from identity-ui's identity-provider form (drag-and-drop or paste) — long-term this belongs in `@scality/core-ui` so both apps share it.

Deviation from the plan: F5 was originally scoped to just a Configure destination step; scope was widened during review to also cover the Source and Replication sections and the Locations-page entry point.

Next: the Apply actions step, which streams the destination-side setup chain and reports its progress.

screenshots :

Screenshot 2026-07-20 at 16 43 06 (2)
Screenshot 2026-07-20 at 16 43 24 (2)

### DIFF
--- a/src/react/locations/CRRSetupWizard/steps/ConfigureStep/ConfigureStep.tsx
+++ b/src/react/locations/CRRSetupWizard/steps/ConfigureStep/ConfigureStep.tsx
@@ -68,7 +68,14 @@ export const ConfigureStep = () => {
       'password',
       'certificate',
     ]);
-    if (!valid) return;
+    if (!valid) {
+      showToast({
+        open: true,
+        status: 'error',
+        message: 'Complete the destination fields before checking the connection',
+      });
+      return;
+    }
     const body = toVerifyBody(getValues());
     try {
       await runVerify(body);
--- a/src/react/locations/CRRSetupWizard/steps/ConfigureStep/ConfigureStep.tsx
+++ b/src/react/locations/CRRSetupWizard/steps/ConfigureStep/ConfigureStep.tsx
@@ -1,0 +1,123 @@
+import { Form, Icon, Stack, useToast } from '@scality/core-ui';
+import { useStepper } from '@scality/core-ui/dist/components/steppers/Stepper.component';
+import { Button } from '@scality/core-ui/dist/next';
+import { useBasenameRelativeNavigate } from '@scality/module-federation';
+import { useRef } from 'react';
+import { FormProvider, useForm } from 'react-hook-form';
+import { ServiceError } from '../../api/crrConfiguratorClient';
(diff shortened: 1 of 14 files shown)


src/react/locations/CRRSetupWizard/steps/ConfigureStep/schema.ts
```diff
@@ -0,0 +1,143 @@
...
+    otherwise: Joi.string().allow(''),
+  }),
+  username: Joi.string().required(),
+  password: Joi.string().required(),
+  certificate: Joi.string()
+    .pattern(/-----BEGIN CERTIFICATE-----/)
+    .required()
+    .messages({ 'string.pattern.base': 'Paste a PEM-encoded certificate' }),
+
+  destinationAccountName: Joi.string().min(1).required(),
```
ZiyangLinScality: Can use accountNameValidationSchema in AccountCreate.tsx

src/react/locations/CRRSetupWizard/steps/ConfigureStep/ConfigureStep.tsx
```diff
@@ -0,0 +1,123 @@
...
+    trigger,
+    formState: { isValid },
+  } = formMethods;
+
+  const runVerify = async (body: VerifyRequestBody) => {
+    await verify.mutateAsync(body);
+    lastVerifiedRef.current = JSON.stringify(body);
+  };
+
+  const onCheckConnection = async () => {
```
ZiyangLinScality, line 61: When click "Check Connection" on an empty form silently does nothing.

ZiyangLinScality (approved):
LGTM

PR #1244, opened 2026-07-21
feature(crr): render the Apply Actions step from the setup event stream

hervedombya:
Renders the wizard's second step against the destination configurator's streaming `replication-setups` response, aligned with the ISV Assistant's `Form + Table (Step / Action / Status)` pattern so the user watches each provisioning step move from `Pending...` to `Success` or `Failed` in place.

The chain surfaces the 10 canonical actions the backend emits — from importing the destination certificate to creating the replication rule — with source-side steps (source account, source bucket, location, rule) rendered only when the caller opted into them, and the destination-side IAM chain (account, user, access key, policy, attach policy to user) always present. Labels reuse the exact ISV wording (`Create IAM User`, `Generate Access Key`, `Create Policy`, `Attach Policy to User`) and interpolate the concrete resource name where relevant (`Create Account on Destination: {name}`, `Create Bucket on Source: {name}`).

The step title reads `Configure {destination-instance-name} for Cross-Region Replication` — the instance name comes from the previous step's Verify response when the destination was reached in management-network mode; it falls back to `ARTESCA` otherwise. Continue advances to Summary automatically once every step succeeds, Retry sits on the failed row only and re-runs the whole (idempotent) chain from scratch, Cancel aborts mid-flight, Exit returns to Configure whenever the mutation is not in flight.

Ships behind the CRR feature flag; runs against the existing setup-mutation mocks until F7 swaps them for the real backend.


ZiyangLinScality (approved):
LGTM

PR #1247, opened 2026-07-22
Extract useCreateLocationMutation into a reusable hook

hervedombya:
Extracts the location-creation mutation that was inlined in `LocationEditor` into its own hook so the CRR setup wizard can reuse it for its `create-location` step.

## What
- New `useCreateLocationMutation` hook (in `locations/hooks/`) wrapping the configuration-overlay create call, surfacing a 422 validation response as the parsed problem body.
- `LocationEditor` now consumes the hook instead of an inline mutation.

## Behaviour
Mostly a pure extraction, with one intentional error-handling fix applied to both location mutations: the original inline `.catch` only re-threw on 422 and silently resolved every other error, so a 500 or network failure reported `isSuccess` (and could start a reconcile wait on a create/update that actually failed). Both `useCreateLocationMutation` (extracted) and the adjacent `updateLocationMutation` now re-throw non-422 errors so they surface correctly. Covered by unit tests (success, 422 problem body, non-422 failure); existing `LocationEditor` tests still pass.

## Next
The CRR wizard's Apply Actions orchestrator will use this hook (alongside the backend setup stream) in a follow-up.

### DIFF
--- a/src/react/locations/LocationEditor.tsx
+++ b/src/react/locations/LocationEditor.tsx
@@ -119,6 +119,7 @@ function LocationEditor() {
           if (error.status === 422) {
             throw await error.json();
           }
+          throw error;
         });
     },
   });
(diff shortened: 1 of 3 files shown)


ZiyangLinScality (approved):
LGTM

PR #1250, opened 2026-07-24
Keep CRR Apply Actions row statuses in sync with the running mutations

hervedombya:
The CRR provisioning wizard's Apply Actions table showed every row stuck on "Pending", even after the underlying operations had already succeeded or failed — so the user got no feedback and the run looked frozen.

Each row's status is derived from the mutation instances handed to `useChainedMutations`, but the config holding them was memoized without depending on those instances. React Query returns a new result object on every status change, so the memo kept the original idle instances and the rows never moved off "Pending".

Depending on the mutation instances in that memo — the approach the ISV Apply Actions step already uses — rebuilds the config whenever a status changes, so rows now reflect live Success / Failed / Retry. The chain is not restarted: `useChainedMutations` keys its reset on the step ids, not the array identity.

Verified against a live ARTESCA platform: the rows now update in real time through the run.


ZiyangLinScality (approved):
LGTM

PR #1263, opened 2026-07-27
fix(mcp): drop dead basePath plumbing on S3-operation tools

damiengillesscality:
## TL;DR
Follow-up to [scality/data-browser#406](https://github.com/scality/data-browser/pull/406), which removes the library's automatic post-tool-call navigation (`autoNavigateAfterOperation`). This zenko-ui wrapper computed and injected `basePath` into the S3-operation tool context solely for that mechanism to consume — it's dead now that the mechanism is gone. The `navigateToRoute` branch, which genuinely needs `basePath`, is untouched.

### DIFF
--- a/src/react/mcpTools/createZenkoS3Tools.ts
+++ b/src/react/mcpTools/createZenkoS3Tools.ts
@@ -199,18 +199,6 @@ export function createZenkoS3Tools(
           'mcp-s3',
         );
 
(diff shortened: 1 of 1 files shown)


damiengillesscality:
Closed automatically when its branch was renamed to `bugfix/webmcp-explicit-navigation` (required prefix). Recreated as #1264.

PR #1266, opened 2026-07-28
feature(crr): build the CRR wizard Summary step

hervedombya:
### TL;DR

Adds the final **Summary** step to the CRR provisioning wizard — a recap of the created location, replication rule and buckets, with a **Finish** action that drops the user on the right page.

### Context

Last user-facing step of the CRR setup wizard, after *Configure* and *Apply Actions*. It also settles the naming of the two artefacts the wizard creates on a single key scheme — `location-<account>-<host>` for the location and `replication-<account>` for the replication rule — shared between the mutations that create them and this recap, so the two can never drift.

### Screenshot

CRR wizard Summary step

### Review focus

- 🟡 `steps/ApplyActionsStep/crrLocation.ts` › `buildCRRLocationName` / `buildCRRReplicationRuleId` — the new identifier scheme. The *create-location* and *create-replication-rule* mutations and the Summary all read names from these two helpers, so what gets created and what gets shown stay in lockstep.
- ⚪ `steps/SummaryStep.tsx` › `onFinish` — routes to the new source bucket when a replication rule was created, and to the buckets list otherwise.

### How to test

1. **Data Management → create a CRR configuration**, complete *Configure* and *Apply Actions* with a replication rule.
2. On **Summary**, confirm it lists the location (`location-<dest-account>-<host>`), the rule (`replication-<dest-account>`), and the source & target buckets.
3. Click **Finish** → you land on the new source bucket page.
4. Re-run **without** a replication rule → only the location is shown, and **Finish** lands on the buckets list.

### DIFF
--- a/src/react/locations/CRRSetupWizard/steps/ApplyActionsStep/ApplyActionsStep.tsx
+++ b/src/react/locations/CRRSetupWizard/steps/ApplyActionsStep/ApplyActionsStep.tsx
@@ -192,6 +192,7 @@ export const ApplyActionsStep = (props: Props) => {
           sourceBucketName: sourceBucketName ?? '',
           targetBucketName: props.targetBucketName ?? '',
           locationName,
+          destinationAccountName: destinationAccountName ?? '',
           destinationRoleArn: result.roleArn,
         });
       };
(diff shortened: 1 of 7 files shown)


ZiyangLinScality (approved):
LGTM

PR #1281, opened 2026-08-28
ZKUI-516: align the Accounts screens with the Buckets screens

Cuervino:
Aligns the Data Management Accounts screens with the Buckets ones, as reported in ARTESCA-18059.

**Title color.** `All Accounts` moves from `textSecondary` to `textPrimary`, which is what `All Buckets` already uses in data-browser's `BrowserPageLayout`.

**Account list icon.** `EntityHeader` had stopped rendering the `icon` prop it receives and hardcoded the single wallet, so `MultiAccountsIcon` was built and thrown away on every render, and both levels showed the same glyph. The icon now goes through core-ui's `IconWrapper`, the same way `BrowserPageLayout` renders `Buckets`. Its viewBox is cropped to the artwork, which carried about 14 units of empty space on the sides, and its height set to `1.4em` rather than the `1em` `svg-inline--fa` imposes, since the drawing is not square like a FontAwesome glyph.

This should be brought to CoreUI, but it will be done via another PR. 

**Back navigation.** An account page gets a back arrow to the account list, mirroring the one the object page of a bucket already offers. Its box is reserved at `r40` by `r32`, the measured size of the rendered `2x` arrow, so the title next to it does not shift while the FontAwesome glyph resolves through its dynamic import.

2026-08-28_15h22_49

image

### DIFF
--- a/src/react/account/MultiAccountsIcon.tsx
+++ b/src/react/account/MultiAccountsIcon.tsx
@@ -1,6 +1,23 @@
-export const MultiAccountsIcon = () => {
+// Follows the shape of core-ui's own custom icons (Bucket, Buckets): the fa
+// classes drive the size and the color is inherited from the wrapper.
+//
+// Two deviations, both because the drawing is not square like a FontAwesome
+// glyph: the viewBox is cropped to the artwork (it used to carry ~14 units of
+// empty space on the sides), and the height is set to 1.4em instead of the 1em
+// `svg-inline--fa` imposes, so the pair of wallets fills the circle the way the
+// bucket icons do rather than floating in the middle of it.
+export const MultiAccountsIcon = ({ size, ariaLabel }: { size?: string; ariaLabel?: string }) => {
   return (
-    <svg width="44" height="28" viewBox="0 0 44 28" fill="none" xmlns="http://www.w3.org/2000/svg">
+    <svg
+      viewBox="7 0 30 28"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+      className={`svg-inline--fa ${size ? `fa-${size}` : ''}`}
+      style={{ height: '1.4em' }}
+      role="img"
+      aria-hidden={!ariaLabel}
+      aria-label={ariaLabel}
+    >
       <path
         fillRule="evenodd"
         clipRule="evenodd"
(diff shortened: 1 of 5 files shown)


src/react/account/MultiAccountsIcon.tsx
```diff
@@ -1,6 +1,23 @@
-export const MultiAccountsIcon = () => {
+// Follows the shape of core-ui's own custom icons (Bucket, Buckets): the fa
```
hervedombya: nit: AI generated useless comments

src/react/ui-elements/ArrowNavigation.tsx
```diff
@@ -0,0 +1,39 @@
+import { Icon } from '@scality/core-ui';
+import { Button } from '@scality/core-ui/dist/next';
+import { spacing } from '@scality/core-ui/dist/spacing';
+import { useBasenameRelativeNavigate } from '@scality/module-federation';
+import styled from 'styled-components';
+
+// core-ui's Button with an icon and neither label nor variant is already the
```
hervedombya: nit: AI generated useless comments

src/react/ui-elements/ArrowNavigation.tsx
```diff
@@ -0,0 +1,39 @@
...
+// painted outside the border box, gets its left edge clipped away.
+const ArrowButton = styled(Button)`
+  width: ${spacing.r40};
+  height: ${spacing.r32};
+  padding: 0;
+  margin-left: ${spacing.r4};
+  flex: none;
+`;
+
+export default function ArrowNavigation({ path, label }: { path: string; label: string }) {
```
hervedombya, line 18: We have the same in data-browser : https://github.com/scality/data-browser/blob/main/packages/data-browser-library/src/components/layouts/ArrowNavigation.tsx#L19
It's better to have this component in core-ui with the same same behaviour

Cuervino: Yeah, indeed.
I prefer the new proposal that takes the ghost button, so the account one is the one that we want to have.
I created a ticket to update the Data browser one. 
What needs to be done on the core UI is to introduce the specific use case in the guideline of the ghost icon button.

Cuervino:
@hervedombya both comment blocks are gone in be753fd. I kept three lines on `ArrowNavigation` for the dynamic import sizing and the clipped focus outline, since neither is obvious from the CSS alone. Ready for your approval whenever you have a minute.

hervedombya (approved):
approved, ArrowNavigation to put in core-ui for next time

